PR #19, opened 2024-04-10
Bug/vote recast array expansion

RonTuretzky:
Addressing #18

### DIFF
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -35,7 +35,6 @@ contract YieldDisburser is OwnableUpgradeable {
      *          Public Functions         *
      *
      */
-
     function distributeYield() public {
         (bool _resolved, /* bytes memory _data */ ) = resolveYieldDistribution();
         require(_resolved, "Yield not resolved");
@@ -73,7 +72,6 @@ contract YieldDisburser is OwnableUpgradeable {
      *           View Functions          *
      *
      */
-
     function resolveYieldDistribution() public view returns (bool, bytes memory) {
         uint48 _now = Time.timestamp();
         uint256 balance = (breadToken.balanceOf(address(this)) + breadToken.yieldAccrued());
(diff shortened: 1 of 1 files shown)


src/YieldDisburser.sol
```diff
@@ -35,7 +35,6 @@ contract YieldDisburser is OwnableUpgradeable {
      *          Public Functions         *
      *
      */
-
```
bagelface, line 38: why removing line breaks? it's inconsistent with other comment blocks now

src/YieldDisburser.sol
```diff
@@ -73,7 +72,6 @@ contract YieldDisburser is OwnableUpgradeable {
      *           View Functions          *
      *
      */
-
```
bagelface, line 76: same as above

src/YieldDisburser.sol
RonTuretzky: Mistake , maybe we need to add a linter , do you know how?

src/YieldDisburser.sol
```diff
@@ -128,25 +125,26 @@ contract YieldDisburser is OwnableUpgradeable {
         require(total == 100, "Total must equal 100");
         if (holderToDistribution[holder].length > 0) {
             delete holderToDistribution[holder];
+        } else {
+            breadchainVoters.push(holder);
```
bagelface, line 129: tbh I don't love the design of this logic. It's not very intuitive and seems to do a lot of unnecessary operations. Why are we explicitly reseting `holderToDistribution[holder]` only to reassign it at `holderToDistribution[holder] = percentages;`? Why not just reassign it? Is it so we can "push" the voter to `breadchainVoters` the first time a vote is cast by that address? If so that's incredibly non-intuitive. IMO this should be rewritten

RonTuretzky: I see what you're saying, I'll try to refactor this

src/YieldDisburser.sol
bagelface: if you're using VS code, there is a Solidity linter that should be included in the plugin and the vscode settings in this repo should be configured to autoformat on save.

RonTuretzky: Works!  I will amend this

PR #22, opened 2024-04-10
Refactor yield distribution

bagelface:
Making a PR for this since we've created a separate ticket for tests

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,0 +1,6 @@
+forge-std/=lib/forge-std/src/
+openzeppelin-contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/
+openzeppelin-contracts/=lib/openzeppelin-contracts/
+lib/openzeppelin-contracts:@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/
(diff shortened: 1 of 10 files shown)


remappings.txt
```diff
@@ -1,6 +1,3 @@
 forge-std/=lib/forge-std/src/
 openzeppelin-contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/
 openzeppelin-contracts/=lib/openzeppelin-contracts/
-lib/openzeppelin-contracts:@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/
```
bagelface, line 4: I removed these remappings because they were causing build errors for me and I figured that would be the case for others as well. I can revert and investigate my local setup if it causes issues for others

src/IBreadToken.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.22;
```
bagelface, line 2: this is important because 0.8.22 because "Solidity 0.8.22 introduces an overflow check optimization that automatically generates an unchecked arithmetic increment of the counter of for loops." https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/

PR #23, opened 2024-04-12
Merge dev into main

bagelface:
Merge the changes we've made after the past couple of weeks with main.


src/YieldDisburser.sol
```diff
@@ -1,126 +1,195 @@
...
-}
+pragma solidity ^0.8.22;
+
+import {OwnableUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import {Checkpoints} from "openzeppelin-contracts/contracts/utils/structs/Checkpoints.sol";
+import {Time} from "openzeppelin-contracts/contracts/utils/types/Time.sol";
+
+import {IBreadToken} from "./IBreadToken.sol";
+
+error AlreadyClaimed();
```
kassandraoftroy, line 10: why this error the only one not defined in the contract but defined outside ? just wondering!

RonTuretzky (approved):

Issue #11, opened 2024-03-27
Feature: Events for yield distribution (to be indexed on frontend)


RonTuretzky:
Events for each bread distribution should include voters for each project, distribution percentage 

- [x] Vote distribution percentage for each project
- [x] base amount distributed for each project 
- [x] Voted amount distributed for each project 
- [x] Voters for each project 

@subject026 please add more if there are any others?

RonTuretzky:
@subject026 Would appreciate your review for the PR , and if it meets the frontend requirements

RonTuretzky:
New requirements: 

~~- [ ] Condense the distribution event to one event as listed here~~
https://github.com/BreadchainCoop/breadchain/pull/27#discussion_r1581008906
- [x] Remove summation of voters 
- [x] Introduce event for single vote

RonTuretzky:
Not condensing event for indexing utility

PR #27, opened 2024-04-22
feat/adding events for frontend indexing

RonTuretzky:
Resolving #11

### DIFF
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -18,7 +18,7 @@ contract YieldDisburser is OwnableUpgradeable {
     uint48 public minimumTimeBetweenClaims;
     mapping(address => uint256[]) public holderToDistribution;
 
-    event ProjectDistributionStats(uint256 votedYield, uint256 baseYield, uint256 percentage, address project);
+    event YieldDistributed(uint256 votedYield, uint256 baseYield, uint256 percentage, address indexed project);
     event TotalVotesForProject(uint256 totalVotes, address project);
 
     error EndAfterCurrentBlock();
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -61,17 +61,17 @@ contract YieldDisburser is OwnableUpgradeable {
 
         uint256 halfBalance = breadToken.balanceOf(address(this)) / 2;
         uint256 baseSplit = halfBalance / breadchainProjects.length;
-        uint256 percentage_of_total_vote;
+        uint256 percentageOfTotalVote;
         uint256 votedSplit;
         uint256[] memory votedSplits;
         uint256[] memory percentages;
 
         for (uint256 i; i < breadchainProjects.length; ++i) {
-            percentage_of_total_vote = projectDistributions[i] / totalVotes;
-            votedSplit = percentage_of_total_vote * halfBalance;
+            percentageOfTotalVote = projectDistributions[i] / totalVotes;
+            votedSplit = percentageOfTotalVote * halfBalance;
             breadToken.transfer(breadchainProjects[i], votedSplit + baseSplit);
             votedSplits[i] = votedSplit;
-            percentages[i] = percentage_of_total_vote;
+            percentages[i] = percentageOfTotalVote;
         }
         emit YieldDistributed(votedSplits, baseSplit, percentages, breadchainProjects);
     }
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -54,19 +54,19 @@ contract YieldDisburser is OwnableUpgradeable {
         if (!_resolved) revert YieldNotResolved();
 
         breadToken.claimYield(breadToken.yieldAccrued(), address(this));
-
-        (uint256[] memory projectDistributions, uint256 totalVotes) = _getVotedDistribution(breadchainProjects.length);
+        uint256 breadchainProjectsLength = breadchainProjects.length;
+        (uint256[] memory projectDistributions, uint256 totalVotes) = _getVotedDistribution(breadchainProjectsLength);
         lastClaimedTimestamp = Time.timestamp();
         lastClaimedBlocknumber = Time.blockNumber();
 
         uint256 halfBalance = breadToken.balanceOf(address(this)) / 2;
-        uint256 baseSplit = halfBalance / breadchainProjects.length;
+        uint256 baseSplit = halfBalance / breadchainProjectsLength;
         uint256 percentageOfTotalVote;
         uint256 votedSplit;
-        uint256[] memory votedSplits;
-        uint256[] memory percentages;
+        uint256[] memory votedSplits = new uint256[](breadchainProjectsLength);
+        uint256[] memory percentages = new uint256[](breadchainProjectsLength);
 
-        for (uint256 i; i < breadchainProjects.length; ++i) {
+        for (uint256 i; i < breadchainProjectsLength; ++i) {
             percentageOfTotalVote = projectDistributions[i] / totalVotes;
             votedSplit = percentageOfTotalVote * halfBalance;
             breadToken.transfer(breadchainProjects[i], votedSplit + baseSplit);
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -17,7 +17,7 @@ contract YieldDisburser is OwnableUpgradeable {
     uint48 public minimumTimeBetweenClaims;
     mapping(address => uint256[]) public holderToDistribution;
 
-    event YieldDistributed(uint256 votedYield, uint256 baseYield, uint256 percentage, address indexed project);
+    event YieldDistributed(uint256[] votedYield, uint256 baseYield, uint256[] percentage, address[] project);
     event BreadHolderVoted(address indexed holder, uint256[] percentages, address[] projects);
 
     error EndAfterCurrentBlock();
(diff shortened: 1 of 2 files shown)


src/YieldDisburser.sol
```diff
@@ -155,9 +158,13 @@ contract YieldDisburser is OwnableUpgradeable {
                 uint256 vote = voterPower * voterDistribution[j];
                 projectDistributions[j] += vote;
                 totalVotes += vote;
+                projectVotes[j] += 1;
```
bagelface: Im confused why we'd use this value instead of the `projectDistributions` value for emitting `TotalVotesForProject` event. We want to know how many voters per project and not the vote distribution?

I also think it would be a good idea to just emit the event here when `i == breadchainVoters.length` instead of doing another loop over `projectCount` later on.

src/YieldDisburser.sol
```diff
@@ -18,7 +18,8 @@ contract YieldDisburser is OwnableUpgradeable {
     uint48 public minimumTimeBetweenClaims;
     mapping(address => uint256[]) public holderToDistribution;
 
-    event BaseYieldDistributed(uint256 amount, address project);
+    event ProjectDistributionStats(uint256 votedYield, uint256 baseYield, uint256 percentage, address project);
```
bagelface: Typically you want your events to be a description of an action, not an entity like "stats" or "total". imo "YieldDistributed" would be a better name for this event. On top of that, I'd consider `percentage` to be redundant data to emit if it can be trivially calculated from the other data emitted. I can see the case for it here because it requires context from the other events to calculate, so I could go either way.

bagelface: `address project` field should be "indexed" so it's easier to search for later on

src/YieldDisburser.sol
```diff
@@ -18,7 +18,8 @@ contract YieldDisburser is OwnableUpgradeable {
...
+    event TotalVotesForProject(uint256 totalVotes, address project);
```
bagelface: `address project` field should be "indexed" so it's easier to search for later on

src/YieldDisburser.sol
```diff
@@ -155,9 +158,13 @@ contract YieldDisburser is OwnableUpgradeable {
                 uint256 vote = voterPower * voterDistribution[j];
                 projectDistributions[j] += vote;
                 totalVotes += vote;
+                projectVotes[j] += 1;
             }
             delete holderToDistribution[voter];
         }
+        for (uint256 i; i < projectCount; ++i) {
```
bagelface: Emitting an event in a "get" function is not ideal because it's not technically actually taking any action. Would be best to return data from it and emit the event in the calling function that is actually making storage changes.

bagelface (changes requested):
see comments

RonTuretzky:
Insightful!! Will fix :)

src/YieldDisburser.sol
RonTuretzky: Hmm, thoughts @subject026 ? Do we prefer to show the amount of voters or the total voting power? 

@bagelface I think there is some benefit to show the amount of humans as opposed the the abstract concept of voting power as it's derived here ...

src/YieldDisburser.sol
RonTuretzky: So you could theoretically take the total yield claimed for that distribution and then divide it by the voted yield + base yield to derive the percentage, but I think that it would just be great QoL this way, and also for people looking on a block explorer would make more sense.

RonTuretzky: resolved with 5584d2f

src/YieldDisburser.sol
RonTuretzky: resolved with 5584d2f

src/YieldDisburser.sol
RonTuretzky: This function is state changing, as it resets the voting for bread holders. Maybe the function should be renamed?

src/YieldDisburser.sol
bagelface: I guess. Technically this is just the number of addresses voting, which isn't necessary limited to one per human so that kind of makes it less meaningful.

src/YieldDisburser.sol
bagelface: Maybe I'm missing something, but this function does not do that. The calling function `distributeYield()` does that when it sets a new "last claimed timestamp/block number".

src/YieldDisburser.sol
bagelface: Ok. One thing to keep in mind here is you'e doing integer division, which will always round to the nearest integer. If the frontend is fine with not displaying any precision beyond an integer, then this is fine. Otherwise, I would advocate for off chain calculation of the percentages.

EDIT: Wait, now that I think of it. This will always result in 0, since `projectDistributions[i]` will always be less than `totalVotes`

src/YieldDisburser.sol
subject026: I'm not sure we need to include any vote data with the disbursement event myself.

I'm imagining 2 events, one when a vote is cast and another when yield is disbursed. 

When a vote is cast we can have the vote values and voter's wallet address and when yield is disbursed I think just the amounts paid out paired with the receiving addresses would be fine.

Everything else can be derived from the vote events and calls to the Bread contract as far as I can tell.

subject026: Also i'm wondering whether there is any reason we shouldn't emit just a single event for disbursement passing a 2D array or something. For example if we just did the wallet addresses and amounts like I described above:
```
[
  [0x12345678910, $250],
  [0x10987654321, £500],
  ...
]
```

src/YieldDisburser.sol
subject026: @bagelface On line 163 the vote is deleted from storage after being counted but we might not need to emit an event here anyway (see my other comment)

bagelface: hmm yea ok yea I misread the function. thought that was a memory variable defined earlier in the function. to be honest, i would like to discuss a larger refactor of this contract with you all because to me it's sort of overly complicated

src/YieldDisburser.sol
RonTuretzky: ok, amending this pr to include your suggestions!

src/YieldDisburser.sol
RonTuretzky: > Typically you want your events to be a description of an action, not an entity like "stats" or "total". imo "YieldDistributed" would be a better name for this event. On top of that, I'd consider `percentage` to be redundant data to emit if it can be trivially calculated from the other data emitted. I can see the case for it here because it requires context from the other events to calculate, so I could go either way.

Ok interesting, I think the reason that this is working so far is because here

```
            uint256 votedSplit = ((projectDistributions[i] * halfBalance) / totalVotes);
            breadToken.transfer(breadchainProjects[i], votedSplit + baseSplit);
```
Yield balance times the votes in all our testing has always been higher than the total votes, this is definitely a bug, I'm opening an issue and resolving this convo

src/YieldDisburser.sol
RonTuretzky: @subject026 Not condensing the event because then we lose the benefit of being able to give the entire history of distribution for a project on the project page

RonTuretzky: conversation should be resolved with [ea93ba3](https://github.com/BreadchainCoop/breadchain/pull/27/commits/ea93ba30664d2ab622cf45945893cb67a02cd76b)

src/YieldDisburser.sol
```diff
@@ -55,17 +56,24 @@ contract YieldDisburser is OwnableUpgradeable {
         breadToken.claimYield(breadToken.yieldAccrued(), address(this));
 
         (uint256[] memory projectDistributions, uint256 totalVotes) = _getVotedDistribution(breadchainProjects.length);
-
         lastClaimedTimestamp = Time.timestamp();
         lastClaimedBlocknumber = Time.blockNumber();
 
         uint256 halfBalance = breadToken.balanceOf(address(this)) / 2;
         uint256 baseSplit = halfBalance / breadchainProjects.length;
+        uint256 percentage_of_total_vote;
```
bagelface: we should stick to either camelCase or snake_case. Most of the contract is in camelCase

src/YieldDisburser.sol
```diff
@@ -55,17 +56,24 @@ contract YieldDisburser is OwnableUpgradeable {
...
+        uint256 votedSplit;
+        uint256[] memory votedSplits;
+        uint256[] memory percentages;
 
         for (uint256 i; i < breadchainProjects.length; ++i) {
-            uint256 votedSplit = ((projectDistributions[i] * halfBalance) / totalVotes);
+            percentage_of_total_vote = projectDistributions[i] / totalVotes;
```
bagelface: is this not the same problem with integer division result in 0 every time because `projectDistributions[i]` will always be less than `totalVotes`?

src/YieldDisburser.sol
RonTuretzky: addressed with a50968d

src/YieldDisburser.sol
RonTuretzky: This pr was created before the percision, which is in a different PR. If we can merge this so I will amend this in the other PR, I just don't want to merge this pr to the other one. Leaving a comment there.

bagelface (approved):
🫡

subject026 (approved):
:bread:

Issue #31, opened 2024-04-27
Bug: calculation of votedSplit could result in 0 value if yieldAccrued/2  * voting power < totalVotes


RonTuretzky:
Context: https://github.com/BreadchainCoop/breadchain/pull/27#discussion_r1581819393

RonTuretzky:
- [x] Add percision variable 
- [x]   amend voteSplit calculatio like so 
          -uint256 votedSplit = ((halfBalance) * (projectDistributions[i] *PERCISION  / totalVotes)) / PERCISION;

Issue #29, opened 2024-04-26
Feature: Receiving/storing votes as points rather than percentages


subject026:
We're allowing users to define the split they want with point values for each project eg:

Project A - 1 points -> 25%
Project B - 2 points -> 50%
Project C - 1 points -> 25%

Currently we're requiring votes to be submitted as percentages meaning these points need be converted to % on the front-end before sending the tx.
 
We want to allow users to change/update their vote anytime before the next disbursement so we'd need to convert the % back into points on the front-end. This wouldn't actually be possible in some cases because of the rounding that would need to happen to convert certain combinations of points to %.

The best approach IMO would be to just store the user's vote as points and convert them in the contract when necessary.

Because of the rounding that will need to happen we can look at using using permyriad (/10000) rather than percent so we can allocate the remainder without having as much of an impact on the amounts paid out.

We could consider distributing based just on the rounded values and leave the small amount of bread left over from the remainders in the pot for the next disbursement.

subject026:
@RonTuretzky I know we discussed this a bit last week but I'm not sure whether what you have in mind would solve the problem when someone wants to change their vote.

RonTuretzky:
I think maybe we should allow the users to just use their actual voting points, this would simplify things a lot.

RonTuretzky:
@subject026 Begining to implement this

PR #33, opened 2024-04-27
feat/storing votes as points and adding percision

RonTuretzky:
#29  & #31

- [ ] After #27 is merged, add precision handling for the view methods as mentioned [here](https://github.com/BreadchainCoop/breadchain/pull/27#discussion_r1582244666)

### DIFF
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -66,7 +66,7 @@ contract YieldDisburser is OwnableUpgradeable {
         uint256 baseSplit = halfBalance / breadchainProjects.length;
 
         for (uint256 i; i < breadchainProjects.length; ++i) {
-            uint256 votedSplit = ((projectDistributions[i] * halfBalance) / totalVotes);
+            uint256 votedSplit = (halfBalance) * (projectDistributions[i]  / totalVotes);
             breadToken.transfer(breadchainProjects[i], votedSplit + baseSplit);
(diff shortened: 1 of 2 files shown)


RonTuretzky:
@bagelface , voting & distribution should be permissionless with no access control, that was the original design at least

src/YieldDisburser.sol
```diff
@@ -15,15 +15,18 @@ contract YieldDisburser is OwnableUpgradeable {
     uint48 public lastClaimedTimestamp;
     uint256 public lastClaimedBlocknumber;
     uint48 public minimumTimeBetweenClaims;
+    uint256 public pointsMax;
     mapping(address => uint256[]) public holderToDistribution;
+    mapping(address => uint256) public holderToDistributionTotal;
+    uint256 public constant PERCISION = 1e18;
```
bagelface: should be `PRECISION`

bagelface: also feel like it can just be a var in local scope of the calculation

bagelface:
Going to wait for https://github.com/BreadchainCoop/breadchain/pull/27 to be merged before approving this because of the precision issues pointed out in that PR

RonTuretzky:
> Going to wait for #27 to be merged before approving this because of the precision issues pointed out in that PR

27 was merged and then this one was merged after

Issue #34, opened 2024-04-27
Feature: Anti Sybil measures


RonTuretzky:
Consider adding features such as min bread holdings, max voters for a period, etc

PR #40, opened 2024-04-28
feat/anti sybil measures

RonTuretzky:
#34


subject026 (commented):
I think basing this on voting power would make more sense.

A user should be able to vote if they have acquired some level of voting power in the current round even if their bread balance is zero at the time they submit a vote.

Also I think the var/error names could be more descriptive eg. something like `minRequiredVotingPower` rather than `sybilMinimum`.

RonTuretzky:
> power

The voting power is pretty abstract though, so I'm not sure what the minimum should be 

Having held 5 bread for at least 10 days for example? 

I'll implement this.

subject026:
Yea it probably make sense to make it a dynamic value that the owner can change.

src/YieldDisburser.sol
```diff
@@ -40,6 +47,10 @@ contract YieldDisburser is OwnableUpgradeable {
         for (uint256 i; i < _breadchainProjects.length; ++i) {
             breadchainProjects[i] = _breadchainProjects[i];
         }
+        minVotingAmount = 5; // must hold atleast 5 bread
+        minVotingHoldingDuration = 10 days; // must hold for atleast 10 days
+        minRequiredVotingPower = 1e18 * minVotingAmount * minVotingHoldingDuration / BLOCKTIME; // Holding 10 bread for 5 days , assuming a 5 second block time
+        maxVotes = 1e4;
```
subject026, line 57: Really don't think this should be hard coded as we don't know what the right threshold is atm and it will likely need to be varied.

Better to have getter/setter for the owner to call and I think a single value makes more sense if voting power is a function of balance/time held anyway. Could initialise it as `0` or pass a value in the constructor.

src/YieldDisburser.sol
```diff
@@ -15,6 +15,12 @@ contract YieldDisburser is OwnableUpgradeable {
     uint48 public lastClaimedTimestamp;
     uint256 public lastClaimedBlocknumber;
     uint48 public minimumTimeBetweenClaims;
+    uint256 public minRequiredVotingPower;
+    uint256 public maxVotes;
+    uint256 public currentVotes;
```
subject026, line 20: Maybe I'm missing something but I can't see `maxVotes` being used anywhere.

Also I'm wondering what the `currentVotes` variable is for exactly. It's being incremented when a vote is cast and reset to zero when `distributeYield` is called but there is also a method for the owner manually update the value?

EDIT: added and removed this as I realised you can select multiple lines when adding comments :sweat_smile:

src/YieldDisburser.sol
RonTuretzky: There is a getter\setter, this is the default value for the initialize but can be changed, because the variables are public there is an automatic getter. 

The best practice is to add them as variables for the initialize, I will open an issue to fix this down the line when finalizing the deployment script, so that it will read from a json file that will hold the values that we want to be sent to the initalizer 

Opened #42 , will resolve in a later pr

src/YieldDisburser.sol
RonTuretzky: yes, will utilize this in later pr, merging for now 

the current votes is supposed to revert when voting above the max votes

Issue #38, opened 2024-04-27
Migration: Add bread repo resources to breadchain repo


RonTuretzky:
Consolidation of all resources to reduce clutter

PR #41, opened 2024-04-29
Migration/bread token resources

RonTuretzky:
addressing #38 

Dependant on PR 20 in the bread repo


src/YieldDisburser.sol
```diff
@@ -2,16 +2,16 @@
...
-import {IBreadToken} from "./IBreadToken.sol";
+import {Bread} from "../lib/bread-token-v2/src/Bread.sol";
 
 error AlreadyClaimed();
 
 contract YieldDisburser is OwnableUpgradeable {
     address[] public breadchainProjects;
     address[] public breadchainVoters;
-    IBreadToken public breadToken;
+    Bread public breadToken;
```
bagelface, line 14: I think this should still be an interface contract that we're importing. Otherwise it will include the code of the entire contract, which we dont need

RonTuretzky: it's only going to increase the deployment cost by very little because it's gnosis, but I can open a pr for an interface on the original repo to add an interface

PR #50, opened 2024-05-27
Add Natspec comments

bagelface:
This is a partial refactor that includes Natspec comments. I'll go through the contracts again and do a more thorough refactor of the business logic and variable naming and all that.

One significant change that was made was I added a wrapper contract for testing that allows us to add "god mode" functions that shouldn't be included in the production contract.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ $ anvil
 ### Test 
 
 ```shell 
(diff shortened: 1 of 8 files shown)


RonTuretzky (approved):

subject026 (approved):
LGTM

PR #51, opened 2024-06-12
Voting power refactor : 1 cycle delay 

RonTuretzky:


subject026 (changes requested):
See https://github.com/BreadchainCoop/breadchain/issues/56

I think the problem is that the most recent checkpoint block number could be higher than the block number at end of calculation period.

RonTuretzky:
> See #56
> 
> I think the problem is that the most recent checkpoint block number could be higher than the block number at end of calculation period.

Yea I think I see the problem. The function wasn't designed with this in mind and the test don't cover it. Essentially if you minted after the end of the previous cycle this will fail.

image

RonTuretzky:
@subject026 please pull and try again, I think this should be addressed with d4809e3

Also I added some comments to the code if you would like to read through, should be more understandable now

subject026 (approved):

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
...
-        uint256 interval_voting_power;
+
+        // We are now at a position where the checkpoint is within the interval
+        // Calculate the voting power for the interval
+        intervalEndValue = intervalEnd._value;
+        votingPowerTotal = intervalEndValue * (_end - prevKey);
+        // If there's a single checkpoint in the interval, return the voting power from the interval (including the edge case where the interval ends at the first checkpoint)
+        if (latestCheckpointPos == 0) {
+            return _end == prevKey ? intervalEndValue : votingPowerTotal;
+        }
```
subject026: Here we're finding the first checkpoint created before the end of the interval but it might have been created before the beginning of the interval.

If that's the only checkpoint (position `0`) we're returning too much VP here as the amount that relates to the period before the start of the interval isn't deducted as in the loop below.

subject026 (changes requested):

bagelface:
imo there are way too many comments added to this and it's making the code messy. The code should be mostly self-commenting with exceptions made for bits of logic that are especially convoluted or require context external to the code. Overly simplistic comments are OK for Natspec comments only because they're meant to be parsed and displayed external to the source code like on Etherscan forms.

this is doing way too much
```
        // Check if the start time is before the end time
        if (_start > _end) revert StartMustBeBeforeEnd();
        // Check if the end time is after the current block
        if (_end > Time.blockNumber()) revert EndAfterCurrentBlock();
        // Get the latest checkpoint for the user
        uint32 latestCheckpointPos = BREAD.numCheckpoints(_account);
        // Check if the user has ever held $BREAD
        if (latestCheckpointPos == 0) revert NoCheckpointsForAccount();
        // Check if the user has ever held $BREAD in the interval by comparing the first mint to the end of the interval
        if (BREAD.checkpoints(_account, 0)._key > _end) return 0;
        // Starting to filter out irrelevant checkpoints that are not in the interval
```

src/YieldDisburser.sol
```diff
@@ -106,32 +109,30 @@ contract YieldDisburser is OwnableUpgradeable {
...
         uint256 _minHoldingDuration,
-        uint256 _maxVotes,
         uint256 _maxPoints,
-        uint48 _minTimeBetweenClaims,
+        uint256 _cycleLength,
         uint48 _lastClaimedTimestamp,
         uint256 _lastClaimedBlockNumber,
         uint256 _precision
     ) public initializer {
-        __Ownable_init(msg.sender);
```
bagelface, line 116: why is this being moved to the bottom of the constructor?

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
...
      * @param _end End time of the period to return the voting power for
      * @param _account Address of user to return the voting power for
-     * @return uint256 Voting power of the specified user at the specified period of time
+     * @return votingPowerTotal  Voting power of the specified user at the specified period of time
      */
-    function getVotingPowerForPeriod(uint256 _start, uint256 _end, address _account) external view returns (uint256) {
+    function getVotingPowerForPeriod(uint256 _start, uint256 _end, address _account)
+        external
+        view
+        returns (uint256 votingPowerTotal)
```
bagelface: return variable should be unnamed to be consistent with all the other return variables

src/YieldDisburser.sol
```diff
@@ -247,33 +239,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
-            percentageOfTotalVote = projectDistributions[i] / totalVotes;
-            votedSplit = halfBalance * (projectDistributions[i] * PRECISION / totalVotes) / PRECISION;
+            percentageOfTotalVote = projectDistributions[i] / currentVotes;
+            votedSplit = halfBalance * (projectDistributions[i] * PRECISION / currentVotes) / PRECISION;
             BREAD.transfer(projects[i], votedSplit + baseSplit);
             votedSplits[i] = votedSplit;
             percentages[i] = percentageOfTotalVote;
         }
         _updateBreadchainProjects();
+        delete  currentVotes;
```
bagelface: remove extra space

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
      * @param _start Start time of the period to return the voting power for
      * @param _end End time of the period to return the voting power for
      * @param _account Address of user to return the voting power for
-     * @return uint256 Voting power of the specified user at the specified period of time
+     * @return votingPowerTotal  Voting power of the specified user at the specified period of time
```
bagelface: revert changes as per comment below and be careful not to add unnecessary spaces to comments

src/YieldDisburser.sol
```diff
@@ -467,4 +419,13 @@ contract YieldDisburser is OwnableUpgradeable {
 
         blockTime = _blockTime;
     }
+    /**
+     * @notice Set a new cycle length
+     * @param _cycleLength New cycle length
+     */
+
+    function setCycleLength(uint256 _cycleLength) public onlyOwner {
+        if (_cycleLength == 0) revert MustBeGreaterThanZero();
```
bagelface: line break after "checks" to be consistent with the rest of the contract

src/YieldDisburser.sol
```diff
@@ -106,32 +109,30 @@ contract YieldDisburser is OwnableUpgradeable {
...
         }
-
+        projectDistributions = new uint256[](projectLength);
         blockTime = _blockTime;
+        PRECISION = _precision;
         minVotingAmount = _minVotingAmount;
-        minHoldingDuration = _minHoldingDuration * 1 days;
-        minRequiredVotingPower = ((minVotingAmount * minHoldingDuration) * PRECISION) / blockTime;
-        maxVotes = _maxVotes;
+        minHoldingDuration = _minHoldingDuration * 1 days; // must hold for atleast _minVotingHoldingDuration  days
```
bagelface: Please double check all your comments. They are consistently inconsistent with expected style.

`// Start with a capital letter with no superfluous spacing`

This one should be:

```// Must hold for at least `_minVotingHoldingDuration` days```

File: src/YieldDisburser.sol [hunk 8]
```diff
@@ -106,32 +109,30 @@ contract YieldDisburser is OwnableUpgradeable {
...
+        minRequiredVotingPower = ((minVotingAmount * minHoldingDuration) * PRECISION) / blockTime; // Holding minVotingAmount bread for minVotingHoldingDuration days , assuming a blockTime second block time
```
bagelface: same comment as above. When referring to referring to the bread token we should style like `$BREAD`

src/YieldDisburser.sol
bagelface: tbh this comment and the comment below are redundant because there is already a comment on the variable definition at the top of the contract for both "minHoldingDuration" and "minRequiredVotingPower"

src/YieldDisburser.sol
```diff
@@ -142,50 +143,16 @@ contract YieldDisburser is OwnableUpgradeable {
...
      * @return bool Flag indicating if the yield distribution is able to be claimed
      * @return bytes Function selector used to distribute the yield
      */
     function resolveYieldDistribution() public view returns (bool, bytes memory) {
-        uint48 _now = Time.timestamp();
+        if (currentVotes == 0) revert NoVotesCasted();
         uint256 balance = (BREAD.balanceOf(address(this)) + BREAD.yieldAccrued());
         if (balance < projects.length) revert YieldTooLow(balance);
-        if (_now < lastClaimedTimestamp + minTimeBetweenClaims) {
+        if (block.number < lastClaimedTimestamp + cycleLength) {
```
bagelface: `block.number` is not the same as `block.timestamp` .. currently this will result in the yield always being in the "already claimed" state after it has been claimed the first time.

bagelface:
Is there a reason we're going through the trouble of converting everything between "timestamps" and "block numbers" instead of just using block numbers and approximating the duration of a cycle to be like 518400 blocks (roughly 30 days assuming Gnosis's 5 second block times)?

src/YieldDisburser.sol
```diff
@@ -106,32 +109,30 @@ contract YieldDisburser is OwnableUpgradeable {
...
         uint256 _maxPoints,
-        uint48 _minTimeBetweenClaims,
+        uint256 _cycleLength,
         uint48 _lastClaimedTimestamp,
         uint256 _lastClaimedBlockNumber,
         uint256 _precision
     ) public initializer {
-        __Ownable_init(msg.sender);
-
         BREAD = Bread(breadAddress);
```
bagelface: not sure why this whole section was refactored. it was specifically formatted to be:

contract initializers
"constant" variable initializers
public variable initializers, with projects broken into it's own section because there was a necessary logic block

src/YieldDisburser.sol
```diff
@@ -321,25 +302,6 @@ contract YieldDisburser is OwnableUpgradeable {
      * @return uint256[] Distribution of votes for projects
```
bagelface: this natspec comment block should be removed

src/YieldDisburser.sol
RonTuretzky: resolved with 16d9641

src/YieldDisburser.sol
RonTuretzky: addresed with 16d9641

src/YieldDisburser.sol
RonTuretzky: addressed with a11a89d

src/YieldDisburser.sol
RonTuretzky: I don't understand, this seems to be consistent with your test wrapper from what I can see (1 line break after "}") 🤔🧐

src/YieldDisburser.sol
RonTuretzky: addresed with ee16c01

src/YieldDisburser.sol
RonTuretzky: amended with 2509d6a

src/YieldDisburser.sol
RonTuretzky: addressed with a11a89d

src/YieldDisburser.sol
RonTuretzky: amended with ee16c01

src/YieldDisburser.sol
RonTuretzky: addressed with a11a89d

src/YieldDisburser.sol
RonTuretzky: addressed with 89bee13

src/YieldDisburser.sol
```diff
@@ -106,32 +109,30 @@ contract YieldDisburser is OwnableUpgradeable {
...
+        PRECISION = _precision;
         minVotingAmount = _minVotingAmount;
-        minHoldingDuration = _minHoldingDuration * 1 days;
-        minRequiredVotingPower = ((minVotingAmount * minHoldingDuration) * PRECISION) / blockTime;
-        maxVotes = _maxVotes;
+        minHoldingDuration = _minHoldingDuration * 1 days; // must hold for atleast _minVotingHoldingDuration  days
+        minRequiredVotingPower = ((minVotingAmount * minHoldingDuration) * PRECISION) / blockTime; // Holding minVotingAmount bread for minVotingHoldingDuration days , assuming a blockTime second block time
         maxPoints = _maxPoints;
-        minTimeBetweenClaims = _minTimeBetweenClaims * 1 days;
+        cycleLength = (_cycleLength * 1 days) / blockTime;
```
bagelface: It's not clear what unit "cycleLength" is supposed to be. it's being passed in to the constructor as "days" and then converted here into "blocks". Then on line 155 it's being added to a time stamp...

src/YieldDisburser.sol
```diff
@@ -467,4 +414,14 @@ contract YieldDisburser is OwnableUpgradeable {
 
         blockTime = _blockTime;
```
bagelface: `setBlockTime` serves no purpose in this contract. The variable is only used in the "initializer" as a constant to calculate initial values. This function should definitely be removed and "blockTime" should be removed as a storage variable entirely.

I'd recommend trying to remove it from the calculations entirely and change the approach to explicitly defining the initial values of variables instead of calculating them in the "initailizer". There's not really a reason to not just pre-calculate the "cycleLength" for example. Same goes for "minRequiredVotingPower" where all the other values in the equation are know before the contract is initialized. It makes things a lot simpler, more predictable, and readable.

src/YieldDisburser.sol
```diff
@@ -467,4 +414,14 @@ contract YieldDisburser is OwnableUpgradeable {
...
     }
+    /**
+     * @notice Set a new cycle length
+     * @param _cycleLength New cycle length
+     */
+
```
bagelface: why is there a line break here??

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
...
+        uint256 intervalEndValue;
+        Checkpoints.Checkpoint208 memory intervalEnd;
+        uint48 prevKey;
+
+        // Find the latest checkpoint that is within the interval
+        while (true) {
+            latestCheckpointPos--;
+            intervalEnd = BREAD.checkpoints(_account, latestCheckpointPos);
+            prevKey = intervalEnd._key;
```
bagelface: imo this makes more sense to be called "latestKey". personally i would prefer the adjective "current" over "latest" but that's personal preference

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
...
-                // If the latest checkpoint is exactly at the end of the interval, return the value at that checkpoint
-                return intervalEndValue;
-            } else {
-                return votingPower; // Otherwise, return the voting power calculated above, which is the value at the latest checkpoint multiplied by the length of the interval
+        // Check if the user has ever held $BREAD in the interval by comparing the first mint to the end of the interval
+        if (BREAD.checkpoints(_account, 0)._key > _end) return 0;
+        // Starting to filter out irrelevant checkpoints that are not in the interval
+
+        uint256 intervalEndValue;
+        Checkpoints.Checkpoint208 memory intervalEnd;
```
bagelface: should this not just be called "checkpoint"??

src/YieldDisburser.sol
```diff
@@ -197,45 +164,70 @@ contract YieldDisburser is OwnableUpgradeable {
...
+                break;
             }
         }
-        uint256 interval_voting_power;
+
+        // We are now at a position where the checkpoint is within the interval
+        // Calculate the voting power for the interval
+        intervalEndValue = intervalEnd._value;
+        uint256 votingPowerTotal = intervalEndValue * (_end - prevKey);
+        // If there's a single checkpoint in the interval, return the voting power from the interval (including the edge case where the interval ends at the first checkpoint)
```
bagelface: does it not make sense that an account would get 0 voting power if the "prevKey = _end" because it would imply the checkpoint was created in the first block the interval ends. to prevent this, just change the break condition in the above "while" loop to be `if (prevKey < _end)`

src/YieldDisburser.sol
```diff
@@ -247,33 +239,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
-        (uint256[] memory projectDistributions, uint256 totalVotes) = _commitVotedDistribution();
-        if (totalVotes == 0) {
-            projectDistributions = new uint256[](projectsLength);
-            for (uint256 i; i < projectsLength; ++i) {
-                projectDistributions[i] = 1;
-            }
-            totalVotes = projectsLength;
-        }
 
         lastClaimedTimestamp = Time.timestamp();
```
bagelface: Currently this algorithm isn't sequenced with the cycles at all, so I assume we are going to be tolerant of some small gaps in between the cycles?

src/YieldDisburser.sol
```diff
@@ -282,64 +267,36 @@ contract YieldDisburser is OwnableUpgradeable {
...
-        if (holderToDistribution[_account].length > 0) {
-            delete holderToDistribution[_account];
-        } else {
-            voters.push(_account);
-        }
-        currentVotes++;
-        holderToDistribution[_account] = _points;
         uint256 total;
         for (uint256 i; i < length; ++i) {
             if (_points[i] > maxPoints) revert VotePointsTooLarge();
```
bagelface: Is there no "maximum sum of points" that a user can define? Only a maximum individual point value regardless of how many projects there are?? I'm not following this logic. Are we not asking users to divide their vote into a distribution out of 100% with a specific level of precision? If so, we should ask them for an array of values where the length = the number of projects and all the values add up to 10000 (assuming two decimals of precision).

src/YieldDisburser.sol
```diff
@@ -247,33 +239,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
         uint256 halfBalance = BREAD.balanceOf(address(this)) / 2;
         uint256 baseSplit = halfBalance / projectsLength;
         uint256 percentageOfTotalVote;
         uint256 votedSplit;
         uint256[] memory votedSplits = new uint256[](projectsLength);
         uint256[] memory percentages = new uint256[](projectsLength);
         for (uint256 i; i < projectsLength; ++i) {
-            percentageOfTotalVote = projectDistributions[i] / totalVotes;
-            votedSplit = halfBalance * (projectDistributions[i] * PRECISION / totalVotes) / PRECISION;
+            percentageOfTotalVote = projectDistributions[i] / currentVotes;
```
bagelface: you will not get a percentage like this, you will also get zero if `currentVotes` is a summation of `projectDistributions` because `projectDistributions[i]` is necessarily smaller than `currentVotes` (unless they are equal which will never practically be the case)

src/YieldDisburser.sol
```diff
@@ -247,33 +239,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
+            votedSplit = halfBalance * (projectDistributions[i] * PRECISION / currentVotes) / PRECISION;
             BREAD.transfer(projects[i], votedSplit + baseSplit);
             votedSplits[i] = votedSplit;
             percentages[i] = percentageOfTotalVote;
         }
         _updateBreadchainProjects();
+        delete currentVotes;
+        delete projectDistributions;
+        projectDistributions = new uint256[](projects.length);
         emit YieldDistributed(votedSplits, baseSplit, percentages, projects);
```
bagelface: imo there is no point in calculating the percentages o chain. just emit the total votes here and use `votedSplits` and `currentVotes` (which ideally is renamed to "totalCycleVotes" or something to be more clear) to calculate the values off chain on the frontend.

src/YieldDisburser.sol
RonTuretzky: addressed by 3d25b34

src/YieldDisburser.sol
RonTuretzky: addressed with f8647c36

RonTuretzky:
> imo there are way too many comments added to this and it's making the code messy. The code should be mostly self-commenting with exceptions made for bits of logic that are especially convoluted or require context external to the code. Overly simplistic comments are OK for Natspec comments only because they're meant to be parsed and displayed external to the source code like on Etherscan forms.
> 
> this is doing way too much
> 
> ```
>         // Check if the start time is before the end time
>         if (_start > _end) revert StartMustBeBeforeEnd();
>         // Check if the end time is after the current block
>         if (_end > Time.blockNumber()) revert EndAfterCurrentBlock();
>         // Get the latest checkpoint for the user
>         uint32 latestCheckpointPos = BREAD.numCheckpoints(_account);
>         // Check if the user has ever held $BREAD
>         if (latestCheckpointPos == 0) revert NoCheckpointsForAccount();
>         // Check if the user has ever held $BREAD in the interval by comparing the first mint to the end of the interval
>         if (BREAD.checkpoints(_account, 0)._key > _end) return 0;
>         // Starting to filter out irrelevant checkpoints that are not in the interval
> ```

addressed by 2e9c094

src/YieldDisburser.sol
RonTuretzky: addressed by 2e9c094 and others

src/YieldDisburser.sol
RonTuretzky: removed by 9e6a03a2a1

src/YieldDisburser.sol
RonTuretzky: addressed by 9e6a03a and others

RonTuretzky:
> Is there a reason we're going through the trouble of converting everything between "timestamps" and "block numbers" instead of just using block numbers and approximating the duration of a cycle to be like 518400 blocks (roughly 30 days assuming Gnosis's 5 second block times)?

addressed by 9e6a03a @bagelface

src/YieldDisburser.sol
RonTuretzky: fixed with b367c79 and others

src/YieldDisburser.sol
RonTuretzky: fixed by 9e6a03a

src/YieldDisburser.sol
RonTuretzky: addressed by 9e6a03a

src/YieldDisburser.sol
RonTuretzky: fixed

src/YieldDisburser.sol
RonTuretzky: fixed

src/YieldDisburser.sol
RonTuretzky: fixed

src/YieldDisburser.sol
RonTuretzky: Don't understand what you mean by this?

src/YieldDisburser.sol
RonTuretzky: Basically what we and @subject026  agreed on was that they would be able to send points any way they want up to maxPoints, so all the values can only add up to projectsLength * maxPoints indeed. We do check that the array is projectsLength long. I think it is the same as asking users to divide their vote into a distribution out specific level of precision here is maxPoints essentially. that's the precision for 1 distribution

src/YieldDisburser.sol
RonTuretzky: fixed with adab285

src/YieldDisburser.sol
RonTuretzky: @subject026 did you already integrate this? we can patch this later but I don't want to invoke more frontend changes atm

subject026: No these events aren't being consumed anywhere yet so feel free to change them.

src/YieldDisburser.sol
subject026: Originally users were to be able to change their vote so we wanted to be able to retrieve the original point values they input when voting.

src/YieldDisburser.sol
```diff
@@ -123,6 +123,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
+
+    /**
+     * @notice Return the current voting power of a user
+     * @param _account Address of the user to return the voting power for
+     * @return uint256 The voting power of the user
+     */
+    function getCurrentVotingPower(address _account) public view returns (uint256) {
+        uint256 votingPower =
+            this.getVotingPowerForPeriod(lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return (votingPower < minRequiredVotingPower ? 0 : votingPower);
```
RonTuretzky, line 143:
```suggestion
        return (votingPower < minRequiredVotingPower ? maxPoints : votingPower);
```

bagelface: don't understand this suggestion? why would we return maxPoints?

src/YieldDisburser.sol
bagelface: isn't this effectively forcing you to waste voting power though?

Issue #53, opened 2024-06-12
getProjectDistributions 


RonTuretzky:
- [ ] should return an array of (address,uint256) representing the distribution per project

Issue #52, opened 2024-06-12
getCurrentVotingPower


RonTuretzky:
- [x] should return 0 if not eligible to vote in current voting cycle
- [x] should return voting power if is eligible to vote in current voting cycle

PR #54, opened 2024-06-13
Feat/voting v2 view functions

RonTuretzky:
#52  #53

### DIFF
--- a/src/YieldDisburser.sol
+++ b/src/YieldDisburser.sol
@@ -136,9 +136,29 @@ contract YieldDisburser is OwnableUpgradeable {
     }
 
     /**
+     * @notice Returns the current distribution of voting power for projects
+     * @return address[] The current breadchain projects
+     * @return uint256[] The current distribution of voting power for projects
+     */
+    function getCurrentVotingDistribution() public view returns (address[]memory, uint256[] memory) {
+        return (projects, projectDistributions);
+    }
+
+    /**
+     * @notice Return the current voting power of a user
+     * @return uint256 The voting power of the user
+     */
+    function getCurrentVotingPower(address _account) public view returns (uint256) {
+        uint256 votingPower =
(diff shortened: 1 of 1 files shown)


RonTuretzky:
PR needes to be patched after #55 to reflect grace minimum

subject026 (approved):

src/YieldDisburser.sol
```diff
@@ -135,6 +135,25 @@ contract YieldDisburser is OwnableUpgradeable {
...
+     * @notice Returns the current distribution of voting power for projects
+     * @return address[] The current breadchain projects
+     * @return uint256[] The current distribution of voting power for projects
+     */
+    function getCurrentVotingDistribution() public view returns (address[] memory, uint256[] memory) {
+        return (projects, projectDistributions);
+    }
+
+    /**
+     * @notice Return the current voting power of a user
```
bagelface, line 148: should include a `@param _account` comment here

bagelface (approved):
lgtm just missing one @param comment

src/YieldDisburser.sol
RonTuretzky: amended with e0cfac4

src/YieldDisburser.sol
```diff
@@ -135,6 +135,26 @@ contract YieldDisburser is OwnableUpgradeable {
...
+     * @param _account Address of the user to return the voting power for
+     * @return uint256 The voting power of the user
+     */
+    function getCurrentVotingPower(address _account) public view returns (uint256) {
+        uint256 votingPower =
+            this.getVotingPowerForPeriod(lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return (votingPower < minRequiredVotingPower ? 0 : votingPower);
+    }
```
subject026, line 156: This function will revert if if a user has never held bread eg. no checkpoints, would be better just to return `0`.

I think we can just return `0` in `getVotingPowerForPeriod()` if there are no checkpoints for an account rather than reverting.

PR #55, opened 2024-06-13
Grace minimum voting power for bread holders

RonTuretzky:
This pr allows any bread holder who has not met the minimum requirement for voting to participate with a symbolic amount of voting power


RonTuretzky:
Tests are broken , do not review yet

subject026:
> Tests are broken , do not review yet

Might just need to add the condition to handle where the account has enough VP to vote.

PR #58, opened 2024-06-27
Non-breaking changes cleanup

bagelface:
There are a few things in here that could technically be considered "breaking changes" but only if the frontend has already integrate them, which in most cases I doubt. It includes stuff like:

- Removing the return value from `BelowMinRequiredVotingPower` error. We can easily query the minimum required voting power from the contract, so this serves no purpose and introduces an inconsistency with some errors returning values and some not.
- `resolveYieldDistribution` was changed to not revert, but return return "false" if the yield is not distributable as per thee Powerpool documentation (https://docs.powerpool.finance/powerpool-and-poweragent-network/power-agent/user-guides-and-instructions/i-want-to-automate-my-tasks/job-registration-guide#resolver-job). Related errors were removed.
- `YieldDistributed` event was changes to not return values that can be calculated easily on the frontend, given it has context of how the yield algorithm works. It helps to reduce the clutter in the `distributeYield` function significantly. @subject026 already confirmed this isn't used by the frontend.
- Unused variables `minHoldingDuration`, `minVotingAmount`, & `blockTime` and related setter functions were removed.


src/YieldDisburser.sol
```diff
@@ -18,114 +17,100 @@ import {Bread} from "bread-token/src/Bread.sol";
...
-    // @notice The event emitted when a holder casts a vote
-    event BreadHolderVoted(address indexed holder, uint256[] percentages, address[] projects);
+    event YieldDistributed(uint256 yield, uint256 totalVotes, uint256[] projectDistributions);
 
     // @notice The address of the $BREAD token contract
     Bread public BREAD;
     // @notice The precision to use for calculations
     uint256 public PRECISION;
-    // @notice The minimum blocks between yield distributions
+    // @notice The minimum number blocks between yield distributions
```
RonTuretzky:
```suggestion
    // @notice The minimum number of blocks between yield distributions
```

src/YieldDisburser.sol
```diff
@@ -167,127 +134,130 @@ contract YieldDisburser is OwnableUpgradeable {
...
+        for (uint32 i = _currentCheckpointIndex; i > 0;) {
+            // Latest checkpoint voting power is calculated when initializing `_totalVotingPower`, so we pre-decrement the index here
+            _currentCheckpoint = BREAD.checkpoints(_account, --i);
 
-        // We are now at a position where the checkpoint is within the interval or the one nearest it
-        value = checkpoint._value;
-        lastKey = uint48((lastKey < _start ? _start : lastKey));
-        uint256 votingPowerTotal = value * (_end - lastKey);
-        // If there's a single checkpoint in the interval, return the voting power from the interval
-        if (latestCheckpointPos == 0 || lastKey < _start) return votingPowerTotal;
```
RonTuretzky, line 197: This was omitted but I don't see where it's included again, specifically the lastKey<_start part

bagelface: its redundant because of the ternary statement above `lastKey < _start ? _start : lastKey` which means `lastKey` will always be at least `== _start`

RonTuretzky: Then I got something wrong but essentially if the first checkpoint t is before the start we don't need to continue the function logic but return the voting power, so it should have been if lastKey == _start , return votingPowerTotal 

Does that make sense?

bagelface: yea, but unless im missing something, this loop should still returns the correct value in that case:

```
_totalVotingPower += _currentCheckpoint._value * (_latestKey - _currentCheckpoint._key);
````
it adds the value for the interval here

```
            // At the start of the interval, deduct voting power accrued before the interval and return the total
            if (_currentCheckpoint._key <= _start) {
                _totalVotingPower -= _currentCheckpoint._value * (_start - _currentCheckpoint._key);
                break;
            }
```
and subtracts the same value here

since it's an edge case I think this is fine?

Comment (RonTuretzky) [reply src/YieldDisburser.sol:197]: I believe this is actually the average case, as most users will not be minting bread every month, but minting once and then voting for many cycles. 

We should not be fetching another checkpoint and adding it at all,  and the subtraction there is for the diff between the start and the second closest checkpoint to the interval, not the closest checkpoint to the interval. The diff between the closest checkpoint in the interval to the start is never subtracted.

File: src/YieldDisburser.sol [hunk 3]
```diff
@@ -167,127 +134,130 @@ contract YieldDisburser is OwnableUpgradeable {
...
+            --_currentCheckpointIndex;
+            _currentCheckpoint = BREAD.checkpoints(_account, _currentCheckpointIndex);
+        } while (_currentCheckpoint._key > _end);
 
-        uint256 value;
-        uint48 lastKey = type(uint48).max;
-        Checkpoints.Checkpoint208 memory checkpoint;
+        // Initialize voting power with the latest checkpoint thats within the interval (or nearest to it)
+        uint48 _latestKey = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
+        uint256 _totalVotingPower = _currentCheckpoint._value * (_end - _latestKey);
```
RonTuretzky, line 156:
```suggestion
        uint256 _totalVotingPower = _currentCheckpoint._value * (_end - _latestKey);
        
        if (_latestKey == _start) return _totalVotingPower; 
        
```

src/YieldDisburser.sol
bagelface: This condition is only for the off chance the latest checkpoint in the interval is at the same block as the start of the interval, in which case we don't need to do any calculation outside the initialization of '_votingPower'. From what I can tell, the results are the same with the refactored function. Maybe we're not talking about the same thing?

RonTuretzky: it's if the latest checkpoint is before the beginning of the interval, we basically adjust the lastKey so that it considers this checkpoint as the relevant checkpoint for the whole interval.

RonTuretzky:
```mermaid
sequenceDiagram
User ->> Bread : Mint 10 bread at block 10 
User ->> Bread : Mint 10 bread at block 30
User ->> YieldDistributer : Vote at block 50  
YieldDistributer ->>  YieldDistributer : Distribute cycle of blocks 40 -60
YieldDistributer ->> YieldDistributer : Consider User voting power as 20 bread  * 20 blocks 
```

This is the expected behavior, but due to this implementation it will add the 20 voting power for 20 blocks, but then also add 10 voting power for 20 blocks because it fetches another checkpoint instead of returning the vp of the first one that is before the interval. 

It fetches the checkpoint of block 30 with (key : 30 , value :20) 

The ternary will then change the  key to 40 and the voting power calculated will be 

value * ( _end - _lastKey) == 20 * (60-40) == 400 

However it will then fetch the second checkpoint (key:10 ,value :10 ) and add 

10 * ( 40 - 10 ) == 300

voting power 

Then it will subtract 

10 * (40 -10)  == 300 
voting power

RonTuretzky: Ok so looking over at this , the loop basically corrects this , but we don't have to go into the loop in the first place. Because in this cast , _latestKey == _start , we will subtract what we have added inside the loop exactly always. It's just a bit awkward I guess? but we can keep this

RonTuretzky: resolved with cbb4cd4

Issue #63, opened 2024-07-09
Re-cast vote flow


RonTuretzky:
- [ ] User should be able to send a transaction with new voting parameters
- [ ] Previous voting power allotted should be deallocated 
- [ ] Should be able to do this many times within a cycle 
Because this feature requires to retain a commitment of the voter's voting distribution, this feature opens up the avenue to allow votes to be a commitment for an indefinite amount of cycles , if that's a desired feature @tbsoc 

```mermaid 
sequenceDiagram
BreadHolder ->> YieldDistributor : Cast vote with distribution [x,y,z]
YieldDistributor ->> YieldDistributor : Record voting power distribution and allot voting power to projects
note left of BreadHolder : Voter has changed his mind and would like to change his vote 
BreadHolder ->> YieldDistributor : Cast vote with distribution [a,b,c]
YieldDistributor ->> YieldDistributor : Access recorded voting power and distribution [x,y,z] 
YieldDistributor ->> YieldDistributor : Deallocate [x,y,z] 
YieldDistributor ->> YieldDistributor : Allocate [a,b,c] 
```

PR #66, opened 2024-07-12
Vote recasting

RonTuretzky:
#63

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -249,9 +249,10 @@ contract YieldDistributor is OwnableUpgradeable {
 
         bool _hasVotedInCycle = accountLastVoted[_account] > lastClaimedBlockNumber;
         uint256[] storage _voterDistributions = voterDistributions[_account];
-        if (!_hasVotedInCycle) delete voterDistributions[_account];
-
-        if (!_hasVotedInCycle) currentVotes += _votingPower;
+        if (!_hasVotedInCycle) {
+            delete voterDistributions[_account];
+            currentVotes += _votingPower;
+        }
 
         for (uint256 i; i < _points.length; ++i) {
             if (!_hasVotedInCycle) _voterDistributions.push(0);
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -245,12 +245,12 @@ contract YieldDistributor is OwnableUpgradeable {
             if (_points[i] > maxPoints) revert ExceedsMaxPoints();
             _totalPoints += _points[i];
         }
+        if (_totalPoints == 0) revert ZeroVotePoints();
 
         bool _hasVotedInCycle = accountLastVoted[_account] > lastClaimedBlockNumber;
         uint256[] storage _voterDistributions = voterDistributions[_account];
         // If the user has not voted in the current cycle, initialize the voter distribution array with the correct length
         if (!_hasVotedInCycle) {
-
             delete voterDistributions[_account];
             for (uint256 i; i < _points.length; ++i) {
                 _voterDistributions.push(0);
@@ -263,8 +263,6 @@ contract YieldDistributor is OwnableUpgradeable {
 
             _voterDistributions[i] = _currentProjectDistribution;
         }
-
-        if (_totalPoints == 0) revert ZeroVotePoints();
 
         accountLastVoted[_account] = block.number;
         if (!_hasVotedInCycle) currentVotes += _votingPower;
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -19,8 +19,6 @@ import {Bread} from "bread-token/src/Bread.sol";
 contract YieldDistributor is OwnableUpgradeable {
     // @notice The error emitted when attempting to add a project that is already in the `projects` array
     error AlreadyMemberProject();
-    // @notice the error emitted when attemping to vote in the same cycle twice
-    error AlreadyVotedInCycle();
     // @notice The error emitted when a user attempts to vote without the minimum required voting power
     error BelowMinRequiredVotingPower();
     // @notice The error emitted when attempting to calculate voting power for a period that has not yet ended
@@ -75,6 +73,8 @@ contract YieldDistributor is OwnableUpgradeable {
     uint256[] public projectDistributions;
     // @notice The last block number in which a specified account cast a vote
     mapping(address => uint256) public accountLastVoted;
+    // @notice The voting power allocated to projects by voters in the current cycle
+    mapping(address => uint256[]) voterDistributions;
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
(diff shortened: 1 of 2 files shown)


src/YieldDistributor.sol
```diff
@@ -242,21 +240,36 @@ contract YieldDistributor is OwnableUpgradeable {
     function _castVote(address _account, uint256[] calldata _points, uint256 _votingPower) internal {
         if (_points.length != projects.length) revert IncorrectNumberOfProjects();
 
+        // This loop cannot be combined with the loop below because the total points must be calculated first
+        // to perform the correct calculation for each project
```
bagelface: I think this comment isn't necessary since it's pretty obvious by the usage of `_totalPoints` in the second loop that it can't be used correctly without calculating it first

src/YieldDistributor.sol
```diff
@@ -242,21 +240,36 @@ contract YieldDistributor is OwnableUpgradeable {
...
         for (uint256 i; i < _points.length; ++i) {
-            projectDistributions[i] += ((_points[i] * _votingPower * PRECISION) / _totalPoints) / PRECISION;
+            uint256 _currentProjectDistribution = ((_points[i] * _votingPower * PRECISION) / _totalPoints) / PRECISION;
+            projectDistributions[i] += _currentProjectDistribution;
+            if (_hasVotedInCycle) projectDistributions[i] -= _voterDistributions[i];
+
+            _voterDistributions[i] = _currentProjectDistribution;
         }
 
+        if (_totalPoints == 0) revert ZeroVotePoints();
```
bagelface: why move this revert down here?

src/YieldDistributor.sol
```diff
@@ -242,21 +240,36 @@ contract YieldDistributor is OwnableUpgradeable {
...
         uint256 _totalPoints;
-
         for (uint256 i; i < _points.length; ++i) {
             if (_points[i] > maxPoints) revert ExceedsMaxPoints();
             _totalPoints += _points[i];
         }
 
-        if (_totalPoints == 0) revert ZeroVotePoints();
+        bool _hasVotedInCycle = accountLastVoted[_account] > lastClaimedBlockNumber;
+        uint256[] storage _voterDistributions = voterDistributions[_account];
```
bagelface, line 252: It doesn't look like `voterDistributions[_account]` ever gets reassigned.

src/YieldDistributor.sol
```diff
@@ -242,21 +240,36 @@ contract YieldDistributor is OwnableUpgradeable {
...
+        // If the user has not voted in the current cycle, initialize the voter distribution array with the correct length
+        if (!_hasVotedInCycle) {
 
+            delete voterDistributions[_account];
```
bagelface, line 253: It doesn't look like `voterDistributions` ever gets reassigned

bagelface (changes requested):
I'd personally like to see this refactored a bit so the "reset" logic of undoing the previous vote distributions is more grouped together. Right now I see 3 iterations over `_points` and I think it can be reduced to two.

src/YieldDistributor.sol
RonTuretzky: addressed with 9cbb997

src/YieldDistributor.sol
RonTuretzky: relic of mistaken refactoring I did, amended with 720b486

src/YieldDistributor.sol
RonTuretzky: it does in L264 
` _voterDistributions[i] = _currentProjectDistribution;`

RonTuretzky:
> Member

Reduced to two loops (I think we already recognized that there must be at least 2 because the first loop is a pre-requisite for the 2nd) 

Grouped all recast logic to L252-L259

RonTuretzky:
> I'd personally like to see this refactored a bit so the "reset" logic of undoing the previous vote distributions is more grouped together. Right now I see 3 iterations over `_points` and I think it can be reduced to two.

With 9539abf
Reduced to two loops (I think we already recognized that there must be at least 2 because the first loop is a pre-requisite for the 2nd)

Grouped all recast logic to L252-L259

src/YieldDistributor.sol
```diff
@@ -243,20 +241,29 @@ contract YieldDistributor is OwnableUpgradeable {
...
 
+        bool _hasVotedInCycle = accountLastVoted[_account] > lastClaimedBlockNumber;
+        uint256[] storage _voterDistributions = voterDistributions[_account];
+        if (!_hasVotedInCycle) delete voterDistributions[_account];
+
+        if (!_hasVotedInCycle) currentVotes += _votingPower;
+
         for (uint256 i; i < _points.length; ++i) {
-            projectDistributions[i] += ((_points[i] * _votingPower * PRECISION) / _totalPoints) / PRECISION;
+            if (!_hasVotedInCycle) _voterDistributions.push(0);
```
bagelface, line 258: prefer an if...else here

src/YieldDistributor.sol
```diff
@@ -243,20 +241,29 @@ contract YieldDistributor is OwnableUpgradeable {
...
         for (uint256 i; i < _points.length; ++i) {
             if (_points[i] > maxPoints) revert ExceedsMaxPoints();
             _totalPoints += _points[i];
         }
-
         if (_totalPoints == 0) revert ZeroVotePoints();
 
+        bool _hasVotedInCycle = accountLastVoted[_account] > lastClaimedBlockNumber;
+        uint256[] storage _voterDistributions = voterDistributions[_account];
+        if (!_hasVotedInCycle) delete voterDistributions[_account];
```
bagelface: combine into a single block

bagelface (changes requested):

src/YieldDistributor.sol
RonTuretzky: addressed with 18a2ddb

src/YieldDistributor.sol
RonTuretzky: addressed with 0d75ed9

test/YieldDistributor.t.sol
```diff
@@ -135,6 +135,38 @@ contract YieldDistributorTest is Test {
...
+        yieldDistributor.castVote(percentages);
+        percentages.pop();
+        percentages.push(70);
+        vm.prank(account);
+        yieldDistributor.castVote(percentages);
+        yieldDistributor.distributeYield();
+
+        // Getting the balance of the project after the distribution and checking if it similiar to the yield accrued (there may be rounding issues)
+        uint256 bread_bal_after = bread.balanceOf(address(this));
+        assertGt(bread_bal_after, yieldAccrued - 3);
```
bagelface: maybe save 3 to a variable call like "tolerance" so it's more clear what's being done here

bagelface (approved):
one optional comment on the tests, but LGTM

test/YieldDistributor.t.sol
bagelface: or "marginOfError". i always try to avoid magic numbers

subject026 (approved):

test/YieldDistributor.t.sol
RonTuretzky: addressed with ed220d9

Issue #45, opened 2024-05-10
Bug: Yield /2  less than breadchainProjectsLength


RonTuretzky:
In the above condition, this would result in the base yield being 0

PR #69, opened 2024-07-23
Fixed distribution setter

RonTuretzky:
#25

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -63,8 +63,6 @@ contract YieldDistributor is OwnableUpgradeable {
     uint256 public lastClaimedBlockNumber;
     // @notice The total number of votes cast in the current cycle
     uint256 public currentVotes;
-    // @notice How much of the yield is divided equally among projects
-    uint256 public yieldFixedSplitDivisor;
     // @notice Array of projects eligible for yield distribution
     address[] public projects;
     // @notice Array of projects queued for addition to the next cycle
@@ -77,6 +75,8 @@ contract YieldDistributor is OwnableUpgradeable {
     mapping(address => uint256) public accountLastVoted;
     // @notice The voting power allocated to projects by voters in the current cycle
     mapping(address => uint256[]) voterDistributions;
+    // @notice How much of the yield is divided equally among projects
+    uint256 public yieldFixedSplitDivisor;
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -186,10 +186,12 @@ contract YieldDistributor is OwnableUpgradeable {
      * @return bytes Calldata used by the resolver to distribute the yield
      */
     function resolveYieldDistribution() public view returns (bool, bytes memory) {
+        uint256 _available_yield = BREAD.balanceOf(address(this)) + BREAD.yieldAccrued();
         if (
             currentVotes == 0 // No votes were cast
                 || block.number < lastClaimedBlockNumber + cycleLength // Already claimed this cycle
-                || BREAD.balanceOf(address(this)) + BREAD.yieldAccrued() < projects.length // Yield is insufficient
+                || _available_yield < projects.length // Yield is insufficient
+                || _available_yield / yieldFixedSplit < projects.length // Yield is insufficient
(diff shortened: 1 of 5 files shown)


src/YieldDistributor.sol
```diff
@@ -182,10 +186,12 @@ contract YieldDistributor is OwnableUpgradeable {
      * @return bytes Calldata used by the resolver to distribute the yield
      */
     function resolveYieldDistribution() public view returns (bool, bytes memory) {
+        uint256 _available_yield = BREAD.balanceOf(address(this)) + BREAD.yieldAccrued();
         if (
             currentVotes == 0 // No votes were cast
                 || block.number < lastClaimedBlockNumber + cycleLength // Already claimed this cycle
-                || BREAD.balanceOf(address(this)) + BREAD.yieldAccrued() < projects.length // Yield is insufficient
+                || _available_yield < projects.length // Yield is insufficient
```
bagelface: this is made redundant by the next line. would remove it for better readability

RonTuretzky: So if yield = 9
projectsLength = 5 
yieldFixedSplit = 2 

it should pass the first line ( 9 < 5 == false) 
but fail the second like ( 9 / 2 < 5 == true) 

This is def an edge case but I'm trying to close #45

bagelface: sorry i think i put my comment on the wrong line. there were two conditions here and one was always smaller than the other one because it was being divided by `yieldFixedSplit` so the one where it wasn't being divided wasn't necessary.

`yield / yieldFixedSplit << yield` because yieldFixedSplit is an unsigned int. and since it's a chain of conditions that are checking to revert, `yield / yieldFixedSplit < projectsLength` should be enough to capture both cases and it should still be able to close #45

src/YieldDistributor.sol
```diff
@@ -63,6 +63,8 @@ contract YieldDistributor is OwnableUpgradeable {
     uint256 public lastClaimedBlockNumber;
     // @notice The total number of votes cast in the current cycle
     uint256 public currentVotes;
+    // @notice How much of the yield is divided equally among projects
```
bagelface: this is not upgrade safe. it will shift the storage location of everything below it. new variables need to be appended to the end of the storage layout. so underneath all existing declarations. maybe add a comment about when/why it was added?

ref: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#modifying-your-contracts

does this get caught by the pipeline to test if a change is upgrade safe?

src/YieldDistributor.sol
```diff
@@ -97,6 +100,7 @@ contract YieldDistributor is OwnableUpgradeable {
         minRequiredVotingPower = _minRequiredVotingPower;
         maxPoints = _maxPoints;
         cycleLength = _cycleLength;
+        yieldFixedSplit = _yieldFixedSplit;
```
bagelface: I don't like the name of this new variable. It overloads the term "split" to mean both "a precise amount of yield" and "the divisor used to calculate the split (previous definition)". I would change this to be `yieldFixedSplitDivisor` so it's more clear

bagelface (changes requested):
changes aren't upgrade safe

src/YieldDistributor.sol
RonTuretzky:
```
    └─ ← [Revert] revert: Upgrade safety validation failed:
✘  src/YieldDistributor.sol:YieldDistributor (upgrades from test/upgrades/v1.0.0/YieldDistributor.sol:YieldDistributor)

      src/YieldDistributor.sol:67: Inserted `yieldFixedSplit`
        > New variables should be placed after all existing inherited variables

FAILED
```

Just merged the upgrade safety validation here and checked, it does fail! 
Will amend :)

src/YieldDistributor.sol
RonTuretzky: addressed with 5a940ad

src/YieldDistributor.sol
RonTuretzky: addressed with 6b024f3

src/YieldDistributor.sol
RonTuretzky: addressed with 50c4082

bagelface (approved):
lgtm

PR #70, opened 2024-07-23
Documentation and requirements for LP voting impl

RonTuretzky:


docs/LP_Voting.md
```diff
@@ -0,0 +1,257 @@
...
+-  LP Token holders should be able to redeem their full amount deposited according to the LP conventions, and lose voting right if they redeem the non-$BREAD option 
+-  Users should maintain usability of the LP frontend 
+-  Removing or adding LP should be supported 
+- Getting voting rights from multiple LP should be supported 
+- Getting voting rights from LP and $BREAD at the same time should be supported 
+## Concessions 
+- Contributing non-$BREAD tokens to the LP will afford voting power 
+- Users will have to alternate between LP Frontend and Breadchain Frontend 
+- Voting power is effective only after `cycleLength` after LP tokens were exchanged for voting rights regardless of duration of tokens being in the LP 
+- The voting token will not be transferable
```
bagelface, line 14: why are they not transferrable?

RonTuretzky: The current implementation assumes that we are tracking each user's deposits , and that $BB is in relation to their deposited assets. 
- Imagine the following scenario: 
	1. I deposit 5 monerium + 5 $BREAD to the LP and receive 10 Monerium Butter $MB 
	3. I deposit the 10 LP tokens and receive 8 $BB
	4. The $BB contract is now in possesion of 10 $MB  and has minted $BB 
	5. I transfer someone my $BB 
	6. They attempt to unwrap the $BB and receive xDAI Butter 
	7. The $BB contract does not have a the xDAI butter and they cannot unwrap their $BB
In order to support transferring $BB we have to then also transfer the underlying assets. This is actually doable if we go with the constant scaling factors, but in the event we elect to go with querying the LP pool for the virtual price then we cannot do so because different users have different amounts of buttered bread for different butters in different points of time. 

Essentially , supporting the ability to transfer the $BB complicates the implementation , and the value added is ... to enable defi for our voting utility token? 

I think that the value added is not worth it and not aligned with our goals for this feature and I the over complication is not worth  essentially. I believe that @kassandraoftroy  may have some more insight on this , I believe there was another reason.

PR #78, opened 2024-08-12
Butter that Bread

### LINKED ISSUE #73: ButteredBread Contract
Referencing #64 
The contract should adhere to the following UML and also inherit OwnableUpgradeable

```mermaid
classDiagram

class IERC20{

totalSupply() external view returns (uint256);

balanceOf(address account) external view returns (uint256);

transfer(address to, uint256 value) external returns (bool);

allowance(address owner, address spender) external view returns (uint256);

approve(address spender, uint256 value) external returns (bool);

transferFrom(address from, address to, uint256 value) external returns (bool);

  

}

class ERC20VotesUpgradeable{

checkpoints(address account, uint32 pos) public view virtual returns (Checkpoints.Checkpoint208 memory)

numCheckpoints(address account) public view virtual returns (uint32)

delegate(address delegatee)

}

  

class ButteredBread {

address => bool whitelistedLP

address => address=> uint256 accountToLPBalance

address => uint256 scalingFactors

deposit(address lp, uint256 amount)

withdraw(address lp,uint256 amount)
mint(address to, uint256 amount)
_burn(address holder, uint256 amount)

  

}

  

ERC20VotesUpgradeable <|-- ButteredBread

  

class ILiquidityPool1{

Curve Monerium <> BREAD

}

class ILiquidityPool2{

Curve xDAI <> BREAD

}

class ILiquidityPool3{

UniswapV3 USDC <> BREAD

}

IERC20<|-- ILiquidityPool1

IERC20<|-- ILiquidityPool2

IERC20<|-- ILiquidityPool3

IERC20<|-- ERC20VotesUpgradeable

  
  
  

ILiquidityPool1 --o ButteredBread

ILiquidityPool2 --o ButteredBread

ILiquidityPool3 --o ButteredBread
```

Reference the function flow here for a step by step , low level perspective 

```mermaid
sequenceDiagram

autonumber

BreadHolder ->> BREAD : Mint 20 BREAD

BreadHolder ->> LP : Deposit $BREAD 10 and recieve n LP Tokens

BreadHolder ->> LP : approve(address,amount)

LP ->> ButteredBread : Approve spending n LP Tokens

BreadHolder ->> ButteredBread : deposit(address,amount)

alt if !whitelistedLP[address]

ButteredBread ->> ButteredBread : Revert

else

ButteredBread ->> LP : transferFrom(address,amount)

ButteredBread ->> ButteredBread : accountToLPBalance[msg.sender][lpaddress]=amount

ButteredBread ->> BreadHolder : mint(msg.sender, amount * scalingFactors[lpaddress])

BreadHolder ->> YieldDistributor : castVote(uin256[] percentages)

YieldDistributor ->> YieldDistributor : voting power = sum([getCurrentVotingPower(token) for token in votingTokens])

YieldDistributor ->> YieldDistributor : _castVote(address _account, uint256[] calldata _points, uint256 _votingPower)

note right of BreadHolder : Some time has passed and BreadHolder wants to withdraw their bread from the LP

BreadHolder ->> ButteredBread : withdraw(address,amount)

alt if !whitelistedLP[address] || accountToLPBalance[msg.sender][lpaddress] < amount

ButteredBread ->> ButteredBread : Revert

else

end

ButteredBread ->> ButteredBread : accountToLPBalance[msg.sender][lpaddress]-=amount

ButteredBread ->> ButteredBread : _burn(msg.sender, amount * scalingFactors[lpaddress])

ButteredBread ->> LP: transfer(msg.sender,amount)

end
```

- [x] All relevant aspects of the UML have been implemented 
- [x] Complementary setters for all variables 
- [x] Unit tests 
- [x] Fuzzy tests
- [x] Override transfer and transferFrom methods to disable their functionality, as this token being transferable will lead to undesired behavior.
- [x] Delegation should always occur on [mint](https://github.com/BreadchainCoop/bread-token-v2/blob/b6d1a575c76b70fd912d69d543aa4a002a2d8f24/src/Bread.sol#L79) as implemented in the $BREAD contract

daopunk:
closes #73


foundry.toml
```diff
@@ -8,8 +8,13 @@ build_info = true
 ffi = true
 ast = true
 extra_output = ["storageLayout"]
+evm_version = "cancun"
+
```
RonTuretzky: What does this serve and what evm version were we using before 😅? The latest?

src/ButteredBread.sol
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @title Breadchain Buttered Bread
+ * @notice Deposit LP tokens (butter) to earn yield
+ * @author Breadchain Collective
+ * @custom:coauthor @RonTuretzky
+ * @custom:coauthor @daopunk
+ */
+contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBread {
+    mapping(address lp => bool allow) public allowlistedLPs;
```
RonTuretzky:
```suggestion
    /// @notice Access control for Breadchain sanctioned liquidity pools
    mapping(address lp => bool allow) public allowlistedLPs;
```

src/ButteredBread.sol
```diff
@@ -0,0 +1,100 @@
...
+    mapping(address lp => uint256 factor) public scalingFactors;
```
RonTuretzky, line 24:
```suggestion
    /// @notice How much ButteredBread should be minted for a Liquidity Pool token (Butter) 
    mapping(address lp => uint256 factor) public scalingFactors;
```

src/ButteredBread.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    /// @notice ButteredBread tokens are non-transferable
+    function transferFrom(address, address, uint256) public virtual override returns (bool) {
+        revert NonTransferable();
+    }
+
+    /// @notice Deposit LP tokens and mint ButteredBread with corresponding LP scaling factor
+    function _deposit(address _account, address _lp, uint256 _amount) internal {
+        uint256 beforeBalance = accountToLPBalances[_account][_lp];
+        accountToLPBalances[_account][_lp] = beforeBalance + _amount;
```
RonTuretzky:
```suggestion
        accountToLPBalances[_account][_lp] += _amount;
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.25;
+
+/**
+ * @title Breadchain Buttered Bread interface
+ */
+interface IButteredBread {
+    error InsufficientFunds();
```
RonTuretzky, line 9:
```suggestion
    /// @notice Occurs when user does not have sufficient Butter to mint ButteredBread
    error InsufficientFunds();
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+    error InvalidValue();
```
RonTuretzky, line 11:
```suggestion
    /// @notice Occurs when an invalid value is attempted to be used in setter functions
    error InvalidValue();
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+    error NotAllowListed();
```
RonTuretzky, line 13:
```suggestion
    /// @notice Occurs when attempting a deposit with a non sanctioned LP
    error NotAllowListed();
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+    error NonTransferable();
```
RonTuretzky, line 15:
```suggestion
    /// @notice Occurs when attempting to transfer ButteredBread , a utility token for voting not meant for trading
    error NonTransferable();
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    event AddButter(address _account, address _lp, uint256 _amount);
```
RonTuretzky, line 20:
```suggestion
    /// @notice Specifics how much LP Token (Butter) has been added  
    event AddButter(address _account, address _lp, uint256 _amount);
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+    event RemoveButter(address _account, address _lp, uint256 _amount);
```
RonTuretzky, line 22:
```suggestion
    /// @notice Specifics how much LP Token (Butter) has been removed
    event RemoveButter(address _account, address _lp, uint256 _amount);
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function initialize(
+        address[] memory _liquidityPools,
+        uint256[] memory _scalingFactors,
+        string memory _name,
+        string memory _symbol
+    ) external;
+
+    function allowlistedLPs(address _lp) external view returns (bool _allowed);
```
RonTuretzky, line 51:
```suggestion
    /// @notice Returns whether a given liquidity pool is Breadchain sanctioned or not 
    function allowlistedLPs(address _lp) external view returns (bool _allowed);
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function scalingFactors(address _lp) external view returns (uint256 _factor);
```
RonTuretzky, line 54:
```suggestion
    /// @notice Returns the factor that determines how much ButteredBread should be minted for a Liquidity Pool token (Butter)
    function scalingFactors(address _lp) external view returns (uint256 _factor);
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function accountToLPBalances(address _account, address _lp) external view returns (uint256 _balance);
```
RonTuretzky:
```suggestion
    /// @notice The amount of LP tokens (Butter) deposited for a an account
    function accountToLPBalances(address _account, address _lp) external view returns (uint256 _balance);
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function deposit(address _lp, uint256 _amount) external;
```
RonTuretzky, line 60:
```suggestion
    /// @notice Deposits Butter (LP Tokens) and mints ButteredBread according to the respective LP scaling factor
    function deposit(address _lp, uint256 _amount) external;
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function withdraw(address _lp, uint256 _amount) external;
```
RonTuretzky, line 63:
```suggestion
    /// @notice Withdraws some amount of Butter (LP token) and burns an amount of the user's ButteredBread according to the respective scaling factor
    function withdraw(address _lp, uint256 _amount) external;
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function modifyAllowList(address _lp, bool _allowed) external;
```
RonTuretzky, line 66:
```suggestion
    /// @notice Defines a liquidity pool's status as sanctioned or unsanctioned by Breadchain
    function modifyAllowList(address _lp, bool _allowed) external;
```

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function modifyScalingFactor(address _lp, uint256 _factor) external;
```
RonTuretzky:
```suggestion
    /// @notice Modifies how much ButteredBread should be minted for a Liquidity Pool token (Butter)
    function modifyScalingFactor(address _lp, uint256 _factor) external;
```

RonTuretzky (changes requested):

src/ButteredBread.sol
```diff
@@ -0,0 +1,100 @@
...
+        if (this.delegates(_account) == address(0)) _delegate(_account, _account);
+
+        emit AddButter(_account, _lp, _amount);
+    }
+
+    /// @notice Withdraw LP tokens and burn ButteredBread with corresponding LP scaling factor
+    function _withdraw(address _account, address _lp, uint256 _amount) internal {
+        uint256 beforeBalance = accountToLPBalances[_account][_lp];
+        if (_amount > beforeBalance) revert InsufficientFunds();
+        accountToLPBalances[_account][_lp] = beforeBalance - _amount;
```
RonTuretzky:
```suggestion
        accountToLPBalances[_account][_lp]  -= _amount;
```

RonTuretzky (changes requested):

foundry.toml
daopunk: Need it to compile the Curve interface to the correct bytecode to interact with the forked pool.

script/Registry.s.sol
bagelface, line 1: does "Registry.s.sol" make sense for this filename? seems like "Constants.s.sol" would make more sense

script/Registry.s.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.25;
+
+// Factories
+address constant CURVE_STABLE_SWAP_FACTORY = 0xbC0797015fcFc47d9C1856639CaE50D0e69FbEE8;
```
bagelface: would be nice to indicate with comments that these contract address are the ones live on Gnosis mainnet

src/ButteredBread.sol
bagelface, line 1: would like to see full natspec comments including comments for "params" and "return" values

src/ButteredBread.sol
```diff
@@ -0,0 +1,103 @@
...
+ */
+contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBread {
+    /// @notice Access control for Breadchain sanctioned liquidity pools
+    mapping(address lp => bool allow) public allowlistedLPs;
+    /// @notice How much ButteredBread should be minted for a Liquidity Pool token (Butter)
+    mapping(address lp => uint256 factor) public scalingFactors;
+    /// @notice Butter balance by account and Liquidity Pool token deposited
+    mapping(address account => mapping(address lp => uint256 balance)) public accountToLPBalances;
+
+    modifier isAllowed(address _lp) {
```
bagelface: change name to `onlyAllowed`

src/ButteredBread.sol
```diff
@@ -0,0 +1,103 @@
...
+        _;
+    }
+
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(
+        address[] memory _liquidityPools,
```
bagelface: believe these can all be "calldata" since they're in an initializer not a constructor

src/ButteredBread.sol
daopunk: i can add those if it's necessary, but will make the counter argument that the contract is quite simple and easy to understand, not to mention many params for functions will be redundant (e.g. deposit, _deposit, withdraw, _withdraw will all have the same param definitions). i understand the use of natspec for more complicated contracts, like setting up governance vars that could be blocks vs timestamps, but don't see the param confusion in regard to this contract.

bagelface: It's not necessarily only for the developers. Some third party sites will use the natspec comments to generate forms to make interacting with the contract more accessible. Etherscan does this. Gnosisscan probably does too. I don't think we'd need the natspec comments on internal functions, only public/external, so redundancy less of a concern.

bagelface, line 1: Commit to either:

`/// @notice Uppercase` or
`/// @notice lowercase`

Other contracts use `// @notice Uppercase` (only two forward slashes) so would prefer to see things kept consistent, but not going to block over it

script/Constants.s.sol
```diff
@@ -0,0 +1,16 @@
...
+
+// Factories
+address constant GNOSIS_CURVE_STABLE_SWAP_FACTORY = 0xbC0797015fcFc47d9C1856639CaE50D0e69FbEE8;
+
+// Liquidity Pools
+address constant GNOSIS_CURVE_POOL_XDAI_BREAD = 0xf3D8F3dE71657D342db60dd714c8a2aE37Eac6B4;
+
+// Tokens
+address constant GNOSIS_BREAD = 0xa555d5344f6FB6c65da19e403Cb4c1eC4a1a5Ee3;
+address constant XDAI = 0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d;
```
bagelface: Probably also add `GNOSIS_` prefix here too. Fwiw, my original comment was mostly just so that it was obvious that these contract values are the ones deployed to Gnosis "mainnet" and not a testnet or local address. `GNOSIS_` prefix works for me.

src/ButteredBread.sol
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @title Breadchain Buttered Bread
+ * @notice Deposit Butter (LP tokens) to earn scaling rewards
+ * @author Breadchain Collective
+ * @custom:coauthor @RonTuretzky
+ * @custom:coauthor @daopunk
+ */
+contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBread {
+    /// @notice Access control for Breadchain sanctioned liquidity pools
+    mapping(address lp => bool allow) public allowlistedLPs;
```
bagelface: think it should be `bool allowed`

src/ButteredBread.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
+    /// @param _initData see IButteredBread
+    function initialize(InitData calldata _initData) external initializer {
+        if (_initData.liquidityPools.length != _initData.scalingFactors.length) revert InvalidValue();
+        __Ownable_init(msg.sender);
```
bagelface, line 42: would suggest some line breaks to keep consistent with other contract like:

```
if (_initData.liquidityPools.length != _initData.scalingFactors.length) revert InvalidValue();

__Ownable_init(msg.sender);
__ERC20_init(_initData.name, _initData.symbol);

for (uint256 i; i < _initData.liquidityPools.length; ++i) {
    allowlistedLPs[_initData.liquidityPools[i]] = true;
    scalingFactors[_initData.liquidityPools[i]] = _initData.scalingFactors[i];
}
```

src/ButteredBread.sol
```diff
@@ -0,0 +1,99 @@
...
+        __ERC20_init(_initData.name, _initData.symbol);
+        for (uint256 i; i < _initData.liquidityPools.length; ++i) {
+            allowlistedLPs[_initData.liquidityPools[i]] = true;
+            scalingFactors[_initData.liquidityPools[i]] = _initData.scalingFactors[i];
+        }
+    }
+
+    /// @notice Deposit LP tokens
+    function deposit(address _lp, uint256 _amount) external virtual onlyAllowed(_lp) {
```
bagelface: do we need these functions to be virtual?

src/ButteredBread.sol
```diff
@@ -0,0 +1,99 @@
...
+        _withdraw(msg.sender, _lp, _amount);
+    }
+
+    /// @notice allow or deny LP token
+    function modifyAllowList(address _lp, bool _allowed) external virtual onlyOwner {
+        allowlistedLPs[_lp] = _allowed;
+    }
+
+    /// @notice set LP token scaling factor
+    function modifyScalingFactor(address _lp, uint256 _factor) external virtual onlyOwner onlyAllowed(_lp) {
```
bagelface: we would need to convert the existing values in `accountToLPBalances` if this value was ever updated, otherwise users will get funds locked because `_burn(_account, _amount * scalingFactors[_lp]);` in `_withdraw` will try to burn more than they have if the scaling factor is decrease. It will also allow users to withdraw more than they put in, if the scaling factor is reduced.

bagelface (changes requested):
there's a critical vulnerability related to scaling factors. see comment

src/ButteredBread.sol
RonTuretzky: So if we change the scaling factor, we should basically do a full redeem of those lp tokens?

RonTuretzky:
Due to the way the voting delay works, syncing this at the time or voting will not configure the voting power for that cycle, only the next one @daopunk ... Not sure if this is a good solution or not . 

I feel maybe when we change the scaling factor, we should submit a list of holders alongside the change and sync them as you've implemented

src/ButteredBread.sol
daopunk: added `_syncVotingWeight` function:

- called upon withdrawal to adjust burn amount
- can be called from governor during cast vote to update user balance
- can be batch called by anyone to update an array of users at one time

daopunk:
> I feel maybe when we change the scaling factor, we should submit a list of holders alongside the change and sync them as you've implemented

Yes, makes sense. Well I think it's still good to update active voters in the next cycle. Plus, there is a batch-sync function that can be called with a list of users. Likewise, they will automatically update during withdrawals of any amount. So it's fairly flexible.

foundry.toml
```diff
@@ -1,3 +1,8 @@
+[fmt]
+line_length = 120
+number_underscore = 'thousands'
+wrap_comments = false
```
RonTuretzky, line 4: @bagelface is this in line with your formatting preferences?

src/ButteredBread.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @notice Set LP token scaling factor
+     * @param _lp Liquidity Pool token
+     * @param _factor Scaling percentage incentive of LP token (e.g. 100 = 1X, 150 = 1.5X, 1000 = 10X)
+     */
+    function modifyScalingFactor(address _lp, uint256 _factor) external onlyOwner onlyAllowed(_lp) {
+        if (_factor < 100) revert InvalidValue();
+        scalingFactors[_lp] = _factor;
+    }
```
RonTuretzky: Can we make `syncVotingWeights` and `syncVotingWeight` internal, and add an address array to this method that gets passed to the `syncVotingWeights` method? This way an administrator is less likely to make a mistake and update the scaling factor without the holders

RonTuretzky (changes requested):
Can you add a deploy script for this contract as well?

src/ButteredBread.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @notice Sync voting weight with LP scaling factor
+     * Note: Can be called from `castVote` on governance contract to update voting weight
+     * @param _account Voting account
+     * @param _lp Liquidity Pool token
+     */
+    function syncVotingWeight(address _account, address _lp) external onlyAllowed(_lp) {
```
bagelface: we should include a public function `syncVotingWeight(address _lp)` that allows a user to sync their own voting weight.

src/interfaces/IButteredBread.sol
bagelface, line 1: Curious to hear why the errors are being defined in the interface instead of the contract where they are actually being used.

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,74 @@
...
+    event RemoveButter(address _account, address _lp, uint256 _amount);
+
+    /**
+     * @param liquidityPools sanctioned LPs
+     * @param scalingFactors scaling factor on mint per sanctioned LP
+     * Note: each scaling factor is a fixed point percent (e.g. 100 = 1X, 150 = 1.5X, 1000 = 10X)
+     * @param name ERC20 token name
+     * @param symbol ERC20 token symbol
+     */
+    struct InitData {
```
bagelface, line 31: Personally would prefer to these to just be individual parameters of the "initialize" function. This sort of creates an unwelcome abstraction imo because it requires the developer to look at the interface for implementation details.

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,74 @@
...
+        string name;
+        string symbol;
+    }
+
+    /**
+     * @param balance Value of deposited Butter (LP tokens)
+     * @param scalingFactor At the time of deposit or updated with `syncVotingWeight` function
+     */
+    struct LPData {
+        uint256 balance;
```
bagelface, line 43: Why are we not just using the ERC20 balance instead? This design imo creates a bit of confusion about what is the "real" token balance of a given account.

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,74 @@
...
+    function initialize(InitData calldata _initData) external;
+
+    /// @notice Returns whether a given liquidity pool is Breadchain sanctioned or not
+    function allowlistedLPs(address _lp) external view returns (bool _allowed);
+
+    /// @notice Returns the factor that determines how much ButteredBread should be minted for a Liquidity Pool token (Butter)
+    function scalingFactors(address _lp) external view returns (uint256 _factor);
+
+    /// @notice The amount of LP tokens (Butter) deposited for a an account
+    function accountToLPData(address _account, address _lp) external view returns (LPData memory _lpData);
```
bagelface: Would rather see this broken out into two functions that return the two pieces of data explicitly like:

`balance(account, lp) returns (uint256)`
`scalingFactor(account, lp) returns (uint256)`

It allows us to remove the interfaces need to have context on "LPData" struct so that it can be moved into the contract definition, which imo makes the contract more readable and easier to work with. `accountToLPData` can just be made public if we want to make it accessible like this.

src/interfaces/IButteredBread.sol
RonTuretzky: I've seen that convention before

bagelface (changes requested):
A few more suggestions, but overall the fix for the scaling factor issue looks good to me.

src/interfaces/IButteredBread.sol
bagelface: Any idea why though? To me its harder to read the contract that way. Not sure what the benefit is. Maybe some people find it "cleaner"?

daopunk: def cleaner, less scrolling, and if needed the errors can be separated into another interface, so multiple interfaces can inherit 1 error interface.

src/interfaces/IButteredBread.sol
daopunk: i mean, anyone used to working with interfaces should be able to easily check that. plus, initializing only ever happens once per contract

src/interfaces/IButteredBread.sol
daopunk: the `balanceOf` refers to BreadedButter, whereas the balance in the LPData struct refers to Butter, which when multiplied with the scaling factor will give the BreadedButter amount corresponding to that LP type.

also, added `accountToLPBalance` to get balance quick instead of having to go into the struct

src/interfaces/IButteredBread.sol
daopunk: we need the LPData struct for math. you can use the `accountToLPBalance` function to view Butter balance

src/ButteredBread.sol
daopunk: they could just put their own address as the parameter, but anyway removed this according to the design update from @RonTuretzky

src/ButteredBread.sol
daopunk: there is an array now. left `syncVotingWeights` as external as a means to manually update the list in case of DDOS, but otherwise is automated

src/interfaces/IButteredBread.sol
RonTuretzky: I think it's a forge docs convention maybe

src/ButteredBread.sol
```diff
@@ -141,7 +141,7 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
         if (_accountToLPData[_account][ZERO_ADDRESS].balance != 1) {
             /// @dev truthy value to check if account has ever made a deposit
             _accountToLPData[_account][ZERO_ADDRESS].balance = 1;
-            depositors.push(_account);
+            _depositors.push(_account);
```
RonTuretzky: missing this logic on the withdraw 

this means that we need to implement logic for checking if the withdraw is a full or partial withdraw request

RonTuretzky: Also this doesn't account for the different kinds of depositors, we need a mapping of the lp address to the `depositors` list  
#80 resolves this issue as well

src/interfaces/IButteredBread.sol
bagelface: It would make sense to me if there were multiple contracts inheriting from the same interface, but this interface is likely only used for calling this contract from another contract without having to import the implementation. In any case, not a hill worth dying on so happy to let it be how it is. Mostly, I just want to enforce consistency across multiple contracts. So for example, if the `YieldDistributor` contract has it's errors defined in the contract instead of the interface, it should be updated either here or there so they are consistent with each other.

RonTuretzky: I can open an issue for that and we'll refactor it isolated from this pr , sounds good?

daopunk: Interfaces are needed simply because referencing interfaces from other contracts is more efficient than bringing the entire contract as a reference into the calling contract. If we anticipate any other contract calling this contract, then we should have an interface.

So, if we need an interface, might as well add efficiency by using it as a place to store events, errors, and structs and inheriting that into the contract. Also, on the topic of structs, they are more memory efficient than using numerous parameters.

Finally, it's fine to throw everything into one file when the protocol is small, but if Breadchain grows into a more complex protocol, it is very helpful to have interfaces as the main code reference when you only care about seeing the functions, structs, events and errors, but not the logic. And yes, I agree it should be a standard.

src/interfaces/IButteredBread.sol
daopunk: also, `public` does not work well with custom struct mappings, hence why i made it internal and added a custom view function

src/ButteredBread.sol
daopunk: yeah, that's why it was just added once ever, despite LP type, so there would not be duplicates. anyway, i merged your PR so this is irrelevant now

src/interfaces/IButteredBread.sol
bagelface: I'm not saying a dev cannot figure out what's going on here, it's just that it creates an unnecessary abstraction of the initializer arguments with not benefit because the struct isn't even packing the data. My suggestion is that any context that can be made available in the implementation contract, should. Vs. abstracting it or tucking it away in the interface. I don't think there's any reason for the interface to be defined in a way that it must be deployed as an "upgradeable" contract. Imo the "initializer" function doesn't need an abstract function in the interface and the "InitData" should just be in the implementation.

src/interfaces/IButteredBread.sol
bagelface: I understand the purpose of an interface and I agree that "ButteredBread" needs one. I'm specifically commenting on the inclusion of the errors in the interface. In my opinion, errors should be "localized" to the implementation of a specific contract for a couple reasons:
- Including them in the interface creates an opportunity where a contract that is calling "ButteredBread" via it's interface throws an error defined in the "ButteredBread" interface in it's own context. Something like that can be a huge headache to debug because it's not something the developer would expect to happen.
- Errors don't really serve a purpose in an interface because they exist specific to the implementation. For example, I create an interface with a "MustBeGreaterThanZero" error, but then I remove that requirement in an upgrade. Now there may be contracts out there that had imported the "ButteredBread" interface with an error that serves no purpose in the updated contract.

foundry.toml
bagelface: tbh I dont really have strong opinions about formatting so long as there is consistency

src/interfaces/IButteredBread.sol
daopunk: It's still useful to have errors in the interface, because custom errors need to be decoded. If a developer is using an interface without errors, that is not ideal, however since the contract inherits the interface, it will have the errors regardless.

As far as "what a developer would expect to happen," that claim is completely subjective, and this is such a well-worn pattern by now that it should be no surprise to any smart contract developer that the errors are in the interface.

I don't understand why there is so much pushback on this particular detail.

src/interfaces/IButteredBread.sol
daopunk: Screenshot 2024-08-27 at 9 45 41 PM

Using a struct on the one-time initiation function looks less abstract than using a bunch of params. But clearly we have a difference in opinion.

src/interfaces/IButteredBread.sol
RonTuretzky: Tabling this discussion for a later date in #85 , lets talk about this later and continue with the current implementation in this PR and refactor it later if needed.

RonTuretzky (approved):

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @title Breadchain Buttered Bread interface
+ */
+interface IButteredBread {
+    /// @notice Occurs when user does not have sufficient Butter to mint ButteredBread
+    error InsufficientFunds();
+    /// @notice Occurs when an invalid value is attempted to be used in setter functions
+    error InvalidValue();
+    /// @notice Occurs when attempting a deposit with a non sanctioned LP
```
bagelface: non-sanctioned

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+    error NotAllowListed();
+    /// @notice Occurs when attempting to transfer ButteredBread , a utility token for voting not meant for trading
```
bagelface: - remove extra comma after "ButteredBread"
- phrasing could be improved. maybe: "a soulbound utility token use for voting"

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+    error NonTransferable();
+    /// @notice Occurs when dependent variable is not set
+    error Unset();
```
bagelface, line 17: `error DependentVariableUnset();`

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    /// @notice Specifics how much LP Token (Butter) has been added
```
bagelface: "Specifies" how much Butter (LP token) has been added

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+    event AddButter(address _account, address _lp, uint256 _amount);
+    /// @notice Specifics how much LP Token (Butter) has been removed
```
bagelface: "Specifies" how much Butter (LP token) has been removed

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+    event RemoveButter(address _account, address _lp, uint256 _amount);
+
+    /**
+     * @param liquidityPools sanctioned LPs
+     * @param scalingFactors scaling factor on mint per sanctioned LP
+     * Note: each scaling factor is a fixed point percent (e.g. 100 = 1X, 150 = 1.5X, 1000 = 10X)
```
bagelface: move to @dev tag at top of the natspec comment block

src/interfaces/IButteredBread.sol
```diff
@@ -0,0 +1,70 @@
...
+    /// @notice Returns the factor that determines how much ButteredBread should be minted for a Liquidity Pool token (Butter)
+    function scalingFactors(address _lp) external view returns (uint256 _factor);
+
+    /// @notice The amount of LP tokens (Butter) deposited for an account
+    function accountToLPBalance(address _account, address _lp) external view returns (uint256 _balance);
+
+    /// @notice Deposits Butter (LP Tokens) and mints ButteredBread according to the respective LP scaling factor
+    function deposit(address _lp, uint256 _amount) external;
+
+    /// @notice Withdraws some amount of Butter (LP token) and burns an amount of the user's ButteredBread according to the respective scaling factor
```
bagelface, line 62: Please stick to a consistent use of "Butter (LP tokens)" or "LP tokens (Butter)". It's mixed throughout the contracts

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+    "openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import {IButteredBread} from "src/interfaces/IButteredBread.sol";
+
+/**
+ * @title Breadchain Buttered Bread
+ * @notice Deposit Butter (LP tokens) to earn scaling rewards
+ * @author Breadchain Collective
+ * @custom:coauthor @RonTuretzky
+ * @custom:coauthor @daopunk
```
bagelface, line 15: For YieldDistributor we added user's who provided feedback to coauthors list, so can you please add @bagelface

  🥹 
👉 👈

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+        if (scalingFactors[_lp] < FIXED_POINT_PERCENT) revert Unset();
+        allowlistedLPs[_lp] = _allowed;
+    }
+
+    /**
+     * @notice Set LP token scaling factor
+     * @param _lp Liquidity Pool token
+     * @param _factor Scaling percentage incentive of LP token (e.g. 100 = 1X, 150 = 1.5X, 1000 = 10X)
+     */
+    function modifyScalingFactor(address _lp, uint256 _factor, address[] calldata holders) external onlyOwner {
```
bagelface: `_holders` to be consistent with the rest of the contract

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+     */
+    function modifyAllowList(address _lp, bool _allowed) external onlyOwner {
+        if (scalingFactors[_lp] < FIXED_POINT_PERCENT) revert Unset();
+        allowlistedLPs[_lp] = _allowed;
+    }
+
+    /**
+     * @notice Set LP token scaling factor
+     * @param _lp Liquidity Pool token
+     * @param _factor Scaling percentage incentive of LP token (e.g. 100 = 1X, 150 = 1.5X, 1000 = 10X)
```
bagelface, line 92: include `_holders` @param natspec comment

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+     */
+    function withdraw(address _lp, uint256 _amount) external onlyAllowed(_lp) {
+        _withdraw(msg.sender, _lp, _amount);
+    }
+
+    /**
+     * @notice Allow or deny LP token
+     * @param _lp Liquidity Pool token
+     * @param _allowed Sanction status of LP token
+     * Note: Must set scaling factor before sanctioning LP token
```
bagelface: Move to @dev tag and put underneath @notice tag

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Access control for Breadchain sanctioned liquidity pools
+    mapping(address lp => bool allowed) public allowlistedLPs;
+    /// @notice How much ButteredBread should be minted for a Liquidity Pool token (Butter)
+    mapping(address lp => uint256 factor) public scalingFactors;
+    /// @notice Butter balance by account and Liquidity Pool token deposited
+    mapping(address account => mapping(address lp => LPData)) internal _accountToLPData;
+
+    modifier onlyAllowed(address _lp) {
+        if (allowlistedLPs[_lp] != true) revert NotAllowListed();
```
bagelface: `if (!allowlistLPs[_lap]) revert NotAllowListed();`

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Deposit LP tokens and mint ButteredBread with corresponding LP scaling factor
+    function _deposit(address _account, address _lp, uint256 _amount) internal {
+        IERC20(_lp).transferFrom(_account, address(this), _amount);
+        _accountToLPData[_account][_lp].balance += _amount;
+
+        uint256 currentScalingFactor = scalingFactors[_lp];
+        _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
+
+        _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
```
bagelface, line 117: `(_amount * currentScalingFactor) / FIXED_POINT_PERCENT` would make order of operations more clear

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Withdraw LP tokens and burn ButteredBread with corresponding LP scaling factor
+    function _withdraw(address _account, address _lp, uint256 _amount) internal {
+        if (_amount > _accountToLPData[_account][_lp].balance) revert InsufficientFunds();
+
+        /// @dev ensure proper accounting in case of admin error in `modifyScalingFactor` where not all holders are updated
+        _syncVotingWeight(_account, _lp);
+        _accountToLPData[_account][_lp].balance -= _amount;
+
+        _burn(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
```
bagelface, line 131: `(_amount * scalingFactors[_lp]) / FIXED_POINT_PERCENT` would make order of operations more clear

src/ButteredBread.sol
```diff
@@ -0,0 +1,168 @@
...
+        IERC20(_lp).transfer(_account, _amount);
+
+        emit RemoveButter(_account, _lp, _amount);
+    }
+
+    function _modifyScalingFactor(address _lp, uint256 _factor, address[] calldata holders) internal {
```
bagelface: `_holders` to be consistent

bagelface (approved):
Functionality looks good. There are a few spelling/grammar and stylistic changes to make, but not going to block on their behalf.

PR #80, opened 2024-08-22
refactor: coupling the holder reset with scaling factor modification

RonTuretzky:


RonTuretzky (commented):
This PR introduces coupling scaling factor modification with vote weight syncing , ensuring the system is synchronised.  Tests were commented out to demonstrate the change easily

This is referencing #78  and #73 

@daopunk @bagelface 

There is also a way to validate that the `holders` list is the real `holders `list by using an aggregative hash.

src/ButteredBread.sol
```diff
@@ -103,28 +90,11 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
...
-     * Note: In case of DDOS, set _sync to false and manually sync with syncVotingWeights(_depositors(), _lp)
      */
-    function modifyScalingFactor(address _lp, uint256 _factor, bool _sync) external onlyOwner {
-        _modifyScalingFactor(_lp, _factor);
-        if (_sync) {
-            for (uint256 i = 0; i < _depositors.length; i++) {
-                _syncVotingWeight(_depositors[i], _lp);
-            }
-        }
+    function modifyScalingFactor(address _lp, uint256 _factor, address[] calldata holders) external onlyOwner {
```
bagelface, line 94: `_factor` shouldn't be an argument, it should pull from `scalingFactors` based on `_lp` arugment

daopunk (approved):

PR #91, opened 2024-09-11
Sync Delegation

### LINKED ISSUE #88: $BB delelgation syncing
- [x] after minting $BB, a check to see if the holder is delegated to someone on the $BREAD contract should be performed , if they are delegated then after minting , a delegation to that holder should be performed instead of the self delegation 
- [x] A test should be added to confirm this behavior 
- [x] #90
- [x] The delegation entry point for the $BB contract should be disabled similarly to transfer 

This behavior does not guarantee 100% synced delegation behavior, as that would require the bread contract to call the buttered bread contract, which is an undesirable implementation pattern. This is a compromise, assuming that the user performs the LP deposit after delegation, or that partial withdraws \ new deposits will sync the delegation state.

daopunk:
closes #88

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -6,6 +6,7 @@ import {ERC20VotesUpgradeable} from
     "openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import {IButteredBread} from "src/interfaces/IButteredBread.sol";
+import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
 
 /**
  * @title Breadchain Buttered Bread
@@ -17,6 +18,8 @@ import {IButteredBread} from "src/interfaces/IButteredBread.sol";
  */
 contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBread {
     uint256 public constant FIXED_POINT_PERCENT = 100;
+
+    IERC20Votes public BREAD;
 
     /// @notice Access control for Breadchain sanctioned liquidity pools
     mapping(address lp => bool allowed) public allowlistedLPs;
@@ -38,6 +41,7 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
     /// @param _initData See IButteredBread
     function initialize(InitData calldata _initData) external initializer {
         if (_initData.liquidityPools.length != _initData.scalingFactors.length) revert InvalidValue();
+        BREAD = IERC20Votes(_initData.breadToken);
 
         __Ownable_init(msg.sender);
         __ERC20_init(_initData.name, _initData.symbol);
@@ -106,6 +110,11 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
         revert NonTransferable();
     }
 
+    /// @notice ButteredBread delegation is determined by the BREAD token
+    function delegate(address) public virtual override {
+        revert NonDelegatable();
+    }
+
     /// @notice Deposit LP tokens and mint ButteredBread with corresponding LP scaling factor
     function _deposit(address _account, address _lp, uint256 _amount) internal {
         IERC20(_lp).transferFrom(_account, address(this), _amount);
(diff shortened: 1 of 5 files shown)


src/ButteredBread.sol
```diff
@@ -115,7 +115,8 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
         _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
 
         _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
-        if (this.delegates(_account) == address(0)) _delegate(_account, _account);
+        if (this.delegates(_account) != address(0)) _delegate(_account, this.delegates(_account));
```
RonTuretzky: essentially here we need to check if in the $BREAD contract (forgot to mention that in the issue, we need to add that contract here) the user is delegated to someone. if they are delegated to someone, we need to delegate their $BB balance to that user as well. So if `if (BREAD.delegates(_account) != address(0)) _delegate(_account, BREAD.delegates(_account))`

RonTuretzky (changes requested):

src/ButteredBread.sol
```diff
@@ -115,14 +124,15 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
         _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
 
         _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
-        if (this.delegates(_account) == address(0)) _delegate(_account, _account);
+        _delegate(_account, BREAD.delegates(_account));
```
RonTuretzky:
```suggestion
       address delegatee = BREAD.delegates(_account);
       delegatee == address(0) ?  _delegate(_account,_account) : _delegate(_account,delegatee ) ;
```

src/ButteredBread.sol
```diff
@@ -115,14 +124,15 @@ contract ButteredBread is ERC20VotesUpgradeable, OwnableUpgradeable, IButteredBr
...
 
         emit AddButter(_account, _lp, _amount);
     }
 
     /// @notice Withdraw LP tokens and burn ButteredBread with corresponding LP scaling factor
     function _withdraw(address _account, address _lp, uint256 _amount) internal {
         if (_amount > _accountToLPData[_account][_lp].balance) revert InsufficientFunds();
+        _delegate(_account, BREAD.delegates(_account));
```
RonTuretzky:
```suggestion
        address delegatee = BREAD.delegates(_account);
        delegatee == address(0) ?  _delegate(_account,_account) : _delegate(_account,delegatee ) ;
```

src/interfaces/IERC20Votes.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.25;
+
+import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
```
RonTuretzky, line 4: really silly that oz does not provide an interface for this class , but I see no better option than what you did 

thoughts @bagelface ?

RonTuretzky (changes requested):

src/ButteredBread.sol
RonTuretzky: this snippet could also instead be made into a function that anyone can call , would reduce the amount of code and also allow a bypass for mismatched state

src/interfaces/IERC20Votes.sol
bagelface: seems fine to me

src/interfaces/IButteredBread.sol
```diff
@@ -11,6 +11,8 @@ interface IButteredBread {
     error InvalidValue();
     /// @notice Occurs when attempting a deposit with a non-sanctioned LP
     error NotAllowListed();
+    /// @notice Occurs when attempting to delegate ButteredBread that is determined by BREAD
```
bagelface: I was sort of confused when reading this comment. Would suggest making it more clear. Suggestion:
`Occurs when attempting to delegate ButteredBread. Delegations are set via the BREAD contract`

test/ButteredBread.t.sol
```diff
@@ -403,3 +406,60 @@ contract ButteredBreadTest_Fuzz is ButteredBreadTest {
...
+        assertEq(bb.delegates(BOBBY), DELEGATEE);
+    }
+
+    function testDelegationRevert() public {
+        vm.startPrank(ALICE);
+        bb.deposit(GNOSIS_CURVE_POOL_XDAI_BREAD, TOKEN_AMOUNT / 3);
+
+        assertEq(bb.delegates(ALICE), ALICE);
+
+        vm.expectRevert();
```
bagelface: prefer testing for explicit errors
`vm.expectRevert(abi.encodeWithSelector(IButteredBread.NonDelegatable.selector));`

src/ButteredBread.sol
bagelface: seems reasonable. also can just call the function code in `_deposit` `_withdraw` instead of repeating code

bagelface (approved):

test/ButteredBread.t.sol
daopunk: oh wow, thank you.... i have had so much trouble formatting custom errors in revert statements, which is why i didn't do it. glad to see the correct format

RonTuretzky (approved):

Issue #89, opened 2024-09-09
Create smart contract "style guide"


bagelface:
Create a "style guide" for smart contract development that sets clear expectations for style, formatting, and organization of code submitted to this codebase. The guidelines should be consistent with industry standards.

bagelface:
Will likely just be creating a reference to this style guide and updating the contracts to be consistent with it: https://github.com/bcnmy/biconomy-solidity-style-guide

It's detailed without being super heavy handed and covers everything we wanted to cover with our style guide and more.

rubydusa:
I'm working on an issue and I just wanted to know if there's some extension or plugin or automatic way to format?

bagelface:
> I'm working on an issue and I just wanted to know if there's some extension or plugin or automatic way to format?

You can use the [foundry formatter](https://book.getfoundry.sh/reference/config/formatter) and/or [solhint](https://github.com/protofire/solhint) but since this is a custom style guide, not everything will be enforced.

bagelface:
Completed with https://github.com/BreadchainCoop/breadchain/pull/93

Issue #85, opened 2024-08-29
Error definition location convention


RonTuretzky:
As brought up in #78 [here](https://github.com/BreadchainCoop/breadchain/pull/78#discussion_r1733808094) there are arguments for and against the location of errors (interface vs implementation), a repository wide convention must be decided and refactors executed to enforce the convention.

bagelface:
will likely be resolved with https://github.com/BreadchainCoop/breadchain/issues/89

PR #93, opened 2024-09-16
Apply style guide

bagelface:
The style guide is super detailed and touches on a lot of things. There is a chance I overlooked some things that still need to be applied, but I think this is a good first pass at it. I focused mostly on ButteredBread contract because it hasn't been deployed yet.

This will resolve both https://github.com/BreadchainCoop/breadchain/issues/89 and https://github.com/BreadchainCoop/breadchain/issues/85

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -7,7 +7,7 @@ import {ERC20VotesUpgradeable} from
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 
 import {IButteredBread} from "src/interfaces/IButteredBread.sol";
-import {IBreadToken} from "src/interfaces/IBreadToken.sol";
+import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
 
 /**
  * @title Breadchain Buttered Bread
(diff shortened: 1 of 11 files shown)


src/ButteredBread.sol
```diff
@@ -5,8 +5,9 @@ import {OwnableUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/a
 import {ERC20VotesUpgradeable} from
     "openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+
 import {IButteredBread} from "src/interfaces/IButteredBread.sol";
-import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
+import {IBreadToken} from "src/interfaces/IBreadToken.sol";
```
RonTuretzky: So in the latest commit of the Bread token submodule there is an interface for the bread contract, but it does not include the ERC20Votes aspects of it. Unfortunately, OZ does not provide an interface for ERC20Votes so we must make it ourselves. However, because there is already an interface for the Bread token contract this renaming introduces ambiguity and confusion...

RonTuretzky (changes requested):
Aside from the ambiguity , lgtm!

src/ButteredBread.sol
bagelface: Can you link to where there is already a Bread token interface? I don't see it anywhere. The intention here was to reduce confusion by naming the interface to be consistent with the way that it was being used. Will change it back if it's doing the opposite.

RonTuretzky (approved):

Issue #77, opened 2024-08-10
Integrate $BB in YieldDistributor


RonTuretzky:
Contingent on #73 #75 
- [x] $BB Contract should be added as an ERC20Votes 
- [x] $BB Should be used to derive voting power with a delay in summation with the voting power derived from $BREAD 
- [x] Logic should be implemented [here](https://github.com/BreadchainCoop/breadchain/blob/471c44367d3e436873419b1564ac7538df76bbe1/src/YieldDistributor.sol#L127) with the refactored `getVotingPowerForPeriod` from #75 
- [x]  A setter should be added (it can be assumed that only bread and buttered bread will be used for voting) 
- [x] Unit tests should be added
- [x] Fuzzy tests should be added 
- [x] Upgrade safety validation should be performed according to the readme with the currently deployed version 

As we do not anticipate any voting tokens used, no additional structure for "voting tokens" should be implemented.

PR #94, opened 2024-09-18
Integrate $BB in YieldDistributor 

RonTuretzky:
#77

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -69,7 +69,13 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
         address[] memory _projects
     ) public initializer {
         __Ownable_init(msg.sender);
-
+        if (
+            _bread == address(0) || _butteredBread == address(0) || _precision == 0 || _minRequiredVotingPower == 0
+                || _maxPoints == 0 || _cycleLength == 0 || _yieldFixedSplitDivisor == 0 || _lastClaimedBlockNumber == 0
+                || _projects.length == 0
+        ) {
+            revert MustBeGreaterThanZero();
+        }
         BREAD = Bread(_bread);
         BUTTEREDBREAD = ERC20VotesUpgradeable(_butteredBread);
         PRECISION = _precision;
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -23,8 +23,6 @@ import {IYieldDistributor} from "src/interfaces/IYieldDistributor.sol";
 contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     /// @notice The address of the $BREAD token contract
     Bread public BREAD;
-    /// @notice The address of the $BUTTEREDBREAD token contract
-    ERC20VotesUpgradeable public BUTTEREDBREAD;
     /// @notice The precision to use for calculations
     uint256 public PRECISION;
     /// @notice The minimum number of blocks between yield distributions
@@ -51,6 +49,8 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     mapping(address => uint256[]) voterDistributions;
     /// @notice How much of the yield is divided equally among projects
     uint256 public yieldFixedSplitDivisor;
+    /// @notice The address of the $BUTTEREDBREAD token contract
+    ERC20VotesUpgradeable public BUTTEREDBREAD;
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -53,6 +53,8 @@ contract YieldDistributor is OwnableUpgradeable {
 
     // @notice The address of the $BREAD token contract
     Bread public BREAD;
+    // @notice The address of the $BUTTEREDBREAD token contract
+    ERC20VotesUpgradeable public BUTTEREDBREAD;
     // @notice The precision to use for calculations
     uint256 public PRECISION;
     // @notice The minimum number of blocks between yield distributions
@@ -127,8 +129,12 @@ contract YieldDistributor is OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return
-            this.getVotingPowerForPeriod(BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return this.getVotingPowerForPeriod(
+            BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account
+        )
+            + this.getVotingPowerForPeriod(
+                BUTTEREDBREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account
+            );
     }
 
     /**
@@ -399,5 +405,13 @@ contract YieldDistributor is OwnableUpgradeable {
         if (_yieldFixedSplitDivisor == 0) revert MustBeGreaterThanZero();
 
         yieldFixedSplitDivisor = _yieldFixedSplitDivisor;
+    }
+
+    /**
+     * @notice Set the ButteredBread token contract
+     * @param _butteredBread Address of the ButteredBread token contract
+     */
+    function setButteredBread(address _butteredBread) public onlyOwner {
+        BUTTEREDBREAD = ERC20VotesUpgradeable(_butteredBread);
     }
 }
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -89,6 +89,7 @@ contract YieldDistributor is OwnableUpgradeable {
 
     function initialize(
         address _bread,
+        address _butteredBread,
         uint256 _precision,
         uint256 _minRequiredVotingPower,
         uint256 _maxPoints,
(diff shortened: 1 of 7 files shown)


RonTuretzky:
Probably should make a new deploy script for "from scratch" deploys, however for practical purposes having these deploy scripts separated is useful. 

I elected to not replicate the tests for buttered bread as it is also an ERC20Votes and all that functionality is already tested for $BREAD

src/YieldDistributor.sol
```diff
@@ -97,8 +101,12 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return
-            this.getVotingPowerForPeriod(BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return this.getVotingPowerForPeriod(
```
daopunk, line 111: is the BB contract is not set for some reason, could revert.
could do following if issue is relevant:
```
uint256 _baseVotingPower = getCurrentVotingPower(_account);
uint256 _derivativeVotingPower = (address(BB) != address(0))
       ? this.getVotingPowerForPeriod(BB, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account)
       : 0;

return _baseVotingPower + _derivativeVotingPower;
```

daopunk (approved):
Looks good other than potential non-set revert issue

src/YieldDistributor.sol
```diff
@@ -23,6 +23,8 @@ import {IYieldDistributor} from "src/interfaces/IYieldDistributor.sol";
 contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     /// @notice The address of the $BREAD token contract
     Bread public BREAD;
+    /// @notice The address of the $BUTTEREDBREAD token contract
+    ERC20VotesUpgradeable public BUTTEREDBREAD;
```
bagelface: new storage variables can't be added in the middle, they need to be appended. also should be name should be `BUTTERED_BREAD`

RonTuretzky: addressed with bf4f97e and opened #96 to track out of order variables

src/YieldDistributor.sol
RonTuretzky: addressed with  f909a84

RonTuretzky:
Will be adding a test soon

src/YieldDistributor.sol
```diff
@@ -97,8 +107,12 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return
-            this.getVotingPowerForPeriod(BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return this.getVotingPowerForPeriod(
```
bagelface, line 110: I think it would be better to check if `BUTTERED_BREAD == address(0)` and just return `0` if so. This sort of creates a dependency loop because `ButteredBread.sol` needs to be initialized with the BREAD address and doesn't include a setter for it.

src/YieldDistributor.sol
```diff
@@ -66,8 +69,15 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
-
+        if (
+            _bread == address(0) || _butteredBread == address(0) || _precision == 0 || _minRequiredVotingPower == 0
+                || _maxPoints == 0 || _cycleLength == 0 || _yieldFixedSplitDivisor == 0 || _lastClaimedBlockNumber == 0
+                || _projects.length == 0
+        ) {
+            revert MustBeGreaterThanZero();
+        }
         BREAD = Bread(_bread);
+        BUTTEREDBREAD = ERC20VotesUpgradeable(_butteredBread);
```
bagelface: Change to `BUTTERED_BREAD`. The casing of it in other places implies it is two words and the style guide dictates two words be separated with an underscore: https://github.com/bcnmy/biconomy-solidity-style-guide?tab=readme-ov-file#constants

src/YieldDistributor.sol
```diff
@@ -66,8 +69,15 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
         address[] memory _projects
     ) public initializer {
         __Ownable_init(msg.sender);
-
+        if (
```
bagelface, line 72: I think this is more readable personally
```
if (
    _bread == address(0) ||
    _butteredBread == address(0) ||
    _precision == 0 ||
    _minRequiredVotingPower == 0 ||
    _maxPoints == 0 ||
    _cycleLength == 0 ||
    _yieldFixedSplitDivisor == 0 ||
    _lastClaimedBlockNumber == 0 ||
    _projects.length == 0
) {
    revert MustBeGreaterThanZero();
}
```

src/YieldDistributor.sol
```diff
@@ -66,8 +69,15 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
+            _bread == address(0) || _butteredBread == address(0) || _precision == 0 || _minRequiredVotingPower == 0
+                || _maxPoints == 0 || _cycleLength == 0 || _yieldFixedSplitDivisor == 0 || _lastClaimedBlockNumber == 0
+                || _projects.length == 0
+        ) {
+            revert MustBeGreaterThanZero();
+        }
```
bagelface, line 78: line break

src/YieldDistributor.sol
```diff
@@ -49,6 +49,8 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     mapping(address => uint256[]) voterDistributions;
     /// @notice How much of the yield is divided equally among projects
     uint256 public yieldFixedSplitDivisor;
+    /// @notice The address of the $BUTTEREDBREAD token contract
```
bagelface: `$BUTTEREDBREAD` implies it's the token symbol because of `$` character. Should change this to: 
```
```

src/YieldDistributor.sol
```diff
@@ -97,8 +107,12 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return
-            this.getVotingPowerForPeriod(BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account);
+        return this.getVotingPowerForPeriod(
+            BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account
+        )
+            + this.getVotingPowerForPeriod(
```
bagelface, line 114: Calling this twice is pretty inefficient. Wonder if it's possible to easily refactor `getVotingPowerForPeriod` function to calculate both values in one pass. The arguments are the same, besides the contracts, so maybe it can be made an array.

EDIT: After looking into the function, doesn't seem to be a super easy way to avoid just passing over each contract, but point still stands

bagelface (changes requested):
a few comments

src/YieldDistributor.sol
RonTuretzky: I'm not sure there's a better way to do it... they will likely have different checkpoints and you have to iterate over all of them in the range for both tokens

src/YieldDistributor.sol
RonTuretzky: addressed with 63228eb

src/YieldDistributor.sol
RonTuretzky: the formatter will override this if I attempt to input that ...

RonTuretzky: formatter overrides this.. do you know if there's a comment I can add for it to ignore this?

src/YieldDistributor.sol
RonTuretzky: addressed with 0558896

src/YieldDistributor.sol
RonTuretzky: addressed with 5fe2c80

src/YieldDistributor.sol
bagelface: Aside from changing `line_length`, which i wouldn't recommend, not really. all good

bagelface (approved):
lgtm

test/ButteredBread.t.sol
```diff
@@ -23,6 +28,21 @@ contract ButteredBreadTest is Test {
...
+    uint256 _blocktime = stdJson.readUint(config_data, "._blocktime");
+    uint256 _maxPoints = stdJson.readUint(config_data, "._maxPoints");
+    uint256 _precision = stdJson.readUint(config_data, "._precision");
+    uint256 _minVotingAmount = stdJson.readUint(config_data, "._minVotingAmount");
+    uint256 _cycleLength = stdJson.readUint(config_data, "._cycleLength");
+    uint256 _minHoldingDuration = stdJson.readUint(config_data, "._minHoldingDuration");
+    uint256 _lastClaimedBlockNumber = stdJson.readUint(config_data, "._lastClaimedBlockNumber");
+    uint256 _yieldFixedSplitDivisor = stdJson.readUint(config_data, "._yieldFixedSplitDivisor");
+    // See test/YieldDistributor.t.sol for explanation of these values
+    uint256 _minRequiredVotingPower = stdJson.readUint(config_data, "._minRequiredVotingPower");
```
daopunk, line 45: i recommend in the future, inheriting the deployment script from the Scripts folder into the test setup, which will also inherit the stdJson datas. but for now, this works

RonTuretzky:
Opened #97 to address issues introduced with this PR , merging for now :)

Issue #96, opened 2024-09-18
Re-order variables for v1


RonTuretzky:
Opening rolling issues for variables in the yield distributor which were introduced in an upgrade , making their location inconsistent with the styling/formatting guidelines 

`BUTTEREDBREAD`
`yieldFixedSplitDivisor`
`beforeLastClaimedBlockNumber`

RonTuretzky:
Not relevant anymore given #112

PR #100, opened 2024-09-28
Voting Multipliers

### LINKED ISSUE #71: Voting Multiplier Spec
## Definition 

We define a voting multiplier as the multiplication of existing voting power by increasing factor ( a number bigger than 1).
For example, some user holds 10 $BREAD , and has completed some tasks on guild.xyz , awarding him some NFTs. These could award him a voting multiplier of 10% (1.1), such that his voting power is now effectively equal to holding 11 $BREAD.    

## Motivation 
Reward users for contributing or participating in Breadchain. Create incentives for further activity. Improve the governance process by weighing active and invested user's votes as more meaningful. 
## Multiplier Sources

#### Gnosis Chain 
- A multiplier that has a source of truth on Gnosis chain 
#### Cross Chain 
- A multiplier which has a source of truth on some other blockchain (for practical reasons , an EVM blockchain which is bridgeable to Gnosis chain)
#### Off Chain 
- A multiplier who's source of truth is some web2 application 

## Requirements 
- It should be possible to define a temporary multiplier
- It should be possible to define a permanent multiplier
- It should be possible to define a dynamic multiplier which changes over time according to some variable 
- Multipliers should support the ERC20Votes delegation design pattern 
- It should be possible to define variable sources of truth for multipliers
- It should be possible to add / remove multipliers for the following voting cycle 
- There should be a shared interface for all multipliers 
- It should be possible to edit the source of truth for a multiplier 
- Multipliers should be detected and attributed to a user at the time of voting 
### Concessions 
- For a given multiplier , some action may be required in order to update / attribute a multiplier (such as minting an NFT , submitting proof of some activity or ownership of some asset) 

## High Level Overview 


This overview suggests an implementation approach that relies on NFTs, allowing the usage of existing NFTs sanctioned by the Breadchain community. This approach also asserts that multipliers which require a custom implementation or some offchain integration can conform to an NFT design, providing a homogenous interface to all multipliers. This approach also provides forward compatibility to existing services which already implement NFT which breadchain may choose to adopt or increase usage of (such as [CharacterSheets](https://play.raidguild.org/), [Guild])https://guild.xyz/breadchain) , [Hats](https://www.hatsprotocol.xyz/) etc )
```mermaid 
classDiagram 

class IMultiplier{
	getMultiplyingFactor(address) external view returns (uint256)
	validUntil(address) external view returns (uint256)
}


class INFTMultiplier{
	IERC721 NFTAddress 
	hasNFT(address) returns (bool)
}

class PermanentNFTMultiplier{
	uint256 multiplyingFactor 
	uint256 validity = type(uint256).max
}

class DynamicNFTMultiplier{
	address => uint256 userToFactor
	address => uint256 userToValidity
}

INFTMultiplier --|> IMultiplier 
PermanentNFTMultiplier --|> INFTMultiplier
DynamicNFTMultiplier --|> INFTMultiplier

class ERC721{

name() public view virtual returns (string memory)
symbol() public view virtual returns (string memory)
tokenURI(uint256 tokenId) public view virtual returns (string memory)
} 
class IERC721{
balanceOf(address owner) external view returns (uint256 balance)
ownerOf(uint256 tokenId) external view returns (address owner);

}
class IProveableMultiplier{
	submitActivities(bytes data) external 
	mint(address to, uint256 tokenId) internal
}

IProveableMultiplier --|> ERC721
IProveableMultiplier --|> DynamicNFTMultiplier 
ERC721 --|> IERC721
INFTMultiplier --o IERC721

class IOffChainProveableMultiplier{
	address pullOracle 
	
}
class IOnChainProveableMultiplier{
	address activityContract

}
class ICrossChainProveableMultiplier{
	address bridge 
}
IOffChainProveableMultiplier --|> IProveableMultiplier
IOnChainProveableMultiplier --|> IProveableMultiplier
ICrossChainProveableMultiplier --|> IProveableMultiplier

```

In order to integrate these multipliers into the existing voting infrastructure , the following contract will be deployed and referenced in the YieldDistributor 
```mermaid
classDiagram
class VotingMultipliers{
	Bread BREAD 
	IMultplier[] whitelistedMultipliers 
	getTotalMultipliers(address) external view returns (uint256)
	queueMultiplierAddition(IMultiplier) external 
	queueMultiplierRemoval(IMultiplier) external
}
```


```mermaid 
sequenceDiagram 
autonumber
BreadHolder ->> YieldDistributor : Cast vote 
YieldDistributor ->> YieldDistributor : Get voting power derived from holder's $BREAD and delegated $BREAD
YieldDistributor ->> VotingMultipliers : Loop and apply all existing multipliers of the holder
YieldDistributor ->> YieldDistributor : Apply Voting power 
```



## Overview of Currently Desired Multipliers 

|                                                 | Description                                                                                                 | Multiplier Source          | Suggested Implementation Interface                                                                                     | Brief Implemtation Overview                                                                                                                                                                                                                                          |
| ----------------------------------------------- | ----------------------------------------------------------------------------------------------------------- | -------------------------- | ---------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Providing liquidity                             | Providing liquidity to some pool sanctioned by the Breadchain team containing $BREAD as an underlying asset | Gnosis chain               | DynamicNFTMultiplier                                                                                                   | Create a $BB NFT,  derive the factor from the $BB balance                                                                                                                                                                                                            |
| Voting streaks  / activity                      | Participation in voting of yield distribution cycles over time                                              | Gnosis chain               | IOnChainProveableMultiplier                                                                                            | Add a counter on the yield distributor contract that counts cycles voted in, use that for factor                                                                                                                                                                     |
| On chain activity                               | Participation in partner blockchains                                                                        | Gnosis chain / cross chain | IOnChainProveableMultiplier / ICrossChainProveableMultiplier                                                           | For Gnosis activities, if the activity is proveable on chain only the logic to prove so will need to be implemented in order to mint the multiplier NFT. For other chains, it will need to be bridged in one step, then the multiplier NFT can be minted in another. |
| Recieving salary in $BREAD on a Gnosis Pay card | Receiving salary into a Gnosis Pay card and converting it into $BREAD                                       | Gnosis chain               | IOnChainProveableMultiplier                                                                                            | User will submit the address of their Gnosis pay card multisig, upon successful verification of ownership the factor will be set based on the bread balance                                                                                                          |
| Development Contributions                       | Contributing code / design / architecture / reviews / audits  to develop Breadchain features                | Off chain                  | DynamicNFTMultiplier (if integrating existing platforms) IOffChainProveableMultiplier (If bootstrapping an interface ) | Recommend to check existing solutions that mint NFTs for github contributions and to integrate that                                                                                                                                                                  |
| Research Contributions                          | Contributing research to some Breadchain sanctioned research target                                         | Off chain                  | DynamicNFTMultiplier (if integrating existing platforms) IOffChainProveableMultiplier (If bootstrapping an interface ) | CharacterSheets                                                                                                                                                                                                                                                      |
| Community Organizing                            | Organizing online events, async events, community projects , irl events                                     | Off chain                  | DynamicNFTMultiplier (if integrating existing platforms) IOffChainProveableMultiplier (If bootstrapping an interface ) | CharacterSheets                                                                                                                                                                                                                                                      |
| Community Participation                         | Participating in weekly meetings, events or conferences                                                     | Off chain                  | DynamicNFTMultiplier (if integrating existing platforms) IOffChainProveableMultiplier (If bootstrapping an interface ) | Create an NFT series for the weekly temp checks, factor will be how many they have. Create for other events as needed.                                                                                                                                               |
| Social Media Activity                           | Participating in the Crypto Leftists discord, guilt, following on social media, liking posts over time      | Off chain                  | DynamicNFTMultiplier (if integrating existing platforms) IOffChainProveableMultiplier (If bootstrapping an interface ) | Recommend to find an existing solution that mints NFTs                                                                                                                                                                                                               |
### Delegation

If a user has delegated their voting power on the $BREAD contract to another user, that should also delegate multipliers.

RonTuretzky:
#71 

This is not upgrade-safe due to new variables introduced in an inherited class. Suggesting to couple this PR with #96 and create a brand new YieldDistributor


src/YieldDistributor.sol
```diff
@@ -108,12 +109,13 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return this.getVotingPowerForPeriod(
-            BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account
-        )
-            + this.getVotingPowerForPeriod(
-                BUTTERED_BREAD, lastClaimedBlockNumber - cycleLength, lastClaimedBlockNumber, _account
-            );
+        uint256 lastCycleStart = lastClaimedBlockNumber - cycleLength;
+        uint256 multiplier = this.getTotalMultipliers(_account);
+        multiplier = multiplier == 0 ? 1e18 : multiplier;
```
bagelface: Two issues:
(1) This should be 1 instead of `1e18` otherwise you're going to be increasing the resultant voting power by `e18`.
(2) You should just be able to use to multiplier result directly to multiply the voting power, so this ternary should be part of the `getTotalMultipliers` function.

src/YieldDistributor.sol
```diff
@@ -108,12 +109,13 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
+        uint256 breadVotingPower = this.getVotingPowerForPeriod(BREAD, lastCycleStart, lastClaimedBlockNumber, _account);
+        uint256 butteredBreadVotingPower =
+            this.getVotingPowerForPeriod(BUTTERED_BREAD, lastCycleStart, lastClaimedBlockNumber, _account);
+        return (breadVotingPower + butteredBreadVotingPower) * multiplier / PRECISION;
```
bagelface: I like to use extra parentheses so order of operations is obvious `((breadVotingPower + butteredBreadVotingPower) * multiplier) / PRECISION;`

bagelface: Ok I see what's going on. TBH I don't think `PRECISION` should be shared here since it's being used in two different calculations. I think the `getTotalMultipliers` function should probably return a tuple like `(multiplier, multiplier_precision)` and that should be used here. Ignore my previous comment about using "1 instead of 1e18"

src/YieldDistributor.sol
```diff
@@ -20,7 +21,7 @@ import {IYieldDistributor} from "src/interfaces/IYieldDistributor.sol";
  * @custom:coauthor kassandra.eth
  * @custom:coauthor theblockchainsocialist.eth
  */
-contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
+contract YieldDistributor is IYieldDistributor, OwnableUpgradeable, VotingMultipliers {
```
bagelface, line 25: TBH I'm a little dubious of the upgrade safety here. It would probably be best to implement VotingMultipliers to use [ERC7201](https://eips.ethereum.org/EIPS/eip-7201) storage layouts (like they do with [OwnableUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/723f8cab09cdae1aca9ec9cc1cfa040c2d4b06c1/contracts/access/OwnableUpgradeable.sol#L27)) to avoid storage layout collisions. It seems complicated but it's actually a pretty trivial one time calculation and some boilerplate code for accessing storage variables.

src/VotingMultipliers.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {IVotingMultipliers, IMultiplier} from "src/interfaces/IVotingMultipliers.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+/// @title VotingMultipliers
+/// @notice A contract for managing voting multipliers
+/// @dev Implements IVotingMultipliers interface
+contract VotingMultipliers is OwnableUpgradeable, IVotingMultipliers {
```
bagelface, line 10: So I think this approach of just a single array is fine, but it's quite in inefficient. Consider most users will probably not have any multipliers. This approach requires every user to check every multiplier that was ever created for every calculation. An alternative approach could be:

- a mapping of approved multipliers (instead of the array to iterate over)
- a mapping of user addresses to a list of multipliers they'd like to apply

An explicit update step to a user's list of multipliers would be required (maybe there's a way to work around that in a way that doesn't tradeoff security? maybe let user's set an "autoapply new multipliers" setting?) and then the ecosystem would only apply those ones. It also gives people more control over their setup. This of course comes with the tradeoff of creating a UI for people to add/remove multipliers, but having a UI for this seems like it would actually be a nice thing to have regardless.

src/YieldDistributor.sol
RonTuretzky: I was thinking we should just deploy a new proxy for this version, and use the setter on the bread contract, as this would also help us resolve #96

src/VotingMultipliers.sol
RonTuretzky: Can def change it to mapping > list but I don't think the user should be able to control their multipliers. If it's a non provable multiplier (for example voting streaks) then it should just be activated \ deactivated without their interaction. If it's something they have to prove (like twitter likes for example) then they should have a UI for submitting the proof , but no further than that. I also can't imagine why a user would want to remove a multiplier?

rubydusa: @bagelface why would a UI be needed to add/remove multipliers? Isn't it possible to check in the UI what multipliers from the allowed multipliers the user has and add them automatically?

I also suggest that the original design can be improved by making it possible to query multiple users at onces, thus reducing external calls to multipliers to be proportional to the amount of multipliers, and not a quadratic complexity of m external calls for n users.

Regarding memory concerns (since many users might not have any multipliers at all, returning 1 as a multiplier for most users) it should be noted memory in solidity is not as expensive as it seems:
```
x^2/512 + 3x (x is number of 32 byte words in memory)
```
This means that for 20,000 gas you can expand memory up to ~2500 EVM words, which is 80kb - compared to external calls this cost is negligible

RonTuretzky: I think that this function is only supposed to be called in the context of a single voter voting (that is, it is called within the context of `castVote`). 

But I'm not sure I understand what you mean, are you referring to making 1 function call to receive the total multiplier for all multipliers? Currently within the design, the multipliers are intended to be separate contracts...Do you mean making it so that multipliers are struct instead? That could def improve efficiency as you say....

However some of these multipliers would require adding custom logic for validation and integrations, which makes it inconvenient as we would need to upgrade the contract for each instance of that happening

rubydusa: nvm, for some reason I thought this is called for each user in some function to calculate the total voting power, but on a secong look it seems this is called only when a user votes

Edit: yeah saw your edit now

RonTuretzky: Check out #71  for more details :)

bagelface: > @bagelface why would a UI be needed to add/remove multipliers? Isn't it possible to check in the UI what multipliers from the allowed multipliers the user has and add them automatically?
> 
> I also suggest that the original design can be improved by making it possible to query multiple users at onces, thus reducing external calls to multipliers to be proportional to the amount of multipliers, and not a quadratic complexity of m external calls for n users.
> 
> Regarding memory concerns (since many users might not have any multipliers at all, returning 1 as a multiplier for most users) it should be noted memory in solidity is not as expensive as it seems:
> 
> ```
> x^2/512 + 3x (x is number of 32 byte words in memory)
> ```
> 
> This means that for 20,000 gas you can expand memory up to ~2500 EVM words, which is 80kb - compared to external calls this cost is negligible

If each multiplier is it's own token contract then we need to either iterate over a master list of token contracts for each voting power calculation or maintain a user => multipliers mapping, which would need to be updated explicitly with each new multiplier token minted. We could potentially build that into the multiplier design. OR we could have some kind of explicit "refresh" function.

Really all I'm suggesting is maybe there is a better way than just maintaining a single master list of all available multiplier token contracts and iterating over them every time a calculation for voting power is made because many users will only have one or two multipliers.

src/YieldDistributor.sol
RonTuretzky: addressed with bf324f1

RonTuretzky: addressed with bf324f1

RonTuretzky: so you're saying each multiplier should have it's own precision potentially? can't we just assume they are all 1e18?

src/VotingMultipliers.sol
RonTuretzky: The problem is that then we are forcing the user to perform some action in order to add the multiplier to their list. So for example,lets say we have a community call attendance multiplier, which is awarded as an NFT for someone for attending a community call. 
In this design, the user would not have to do anything other than attend the call, next time he votes during the master list iteriation the YD will detect they are eligible for a voting multiplier and will apply it. 

In the suggested design , the user will have to perform some action (or through account abstraction / refresh function breadchain will have to account for this) in order for the multiplier to be added to their list. 

While this is more gas efficient, the tradeoff here is either offchain infrastructure and additional engineering to maintain a synchronized state , or additional friction on the user. 

Because we are deploying on Gnosis , the actual additional cost on the user to iterate through the master list is negligible, so my opinion is to ignore it. 

The current  cost of voting is $[0.000169425051222655](https://gnosisscan.io/tx/0x0e0e4e0a0051ad4275eeaca6927c9434e58640f260515c632f2552c68a673bc6) and I'm guessing that this iteriation will not increase it in any order of magnitude. 

For reference, users can get 0.0005 / week through through the [Gnosis Faucet](https://faucet.gnosischain.com/) 

@rubydusa would love your opinion on this as well

bagelface: Im not necessarily suggesting any specific design, just pointing out a potentially issue with the current design. I don't think tx costs are a huge concern, but the block gas limits might (30,000,000). We should run a gas simulation on 1,000 and 10,000 multipliers and see how it looks. If it seems like it might be, maybe we can find some tradeoff that automatically pushes an update to a user's list when a new multiplier is issued.

RonTuretzky: Added tests with [5b9b7b3](https://github.com/BreadchainCoop/breadchain/pull/100/commits/5b9b7b3e70b70368530d903761fc1d61d4dd0f00)


|                  | Gas       | $ Cost                |
| ---------------- | --------- | --------------------- |
| Current          | 152_133   | 0.000228199501217064  |
| 100 Multipliers  | 364_702   | 0.000547053002917616  |
| 1000 Multipliers | 2_272_819 | 0.0034092285181825522 |


It seems the increase is not substantial , and I'm doubting we ever get to 1000 multipliers.

RonTuretzky:
Deprioritizing for now

src/YieldDistributor.sol
```diff
@@ -110,8 +111,13 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return this.getVotingPowerForPeriod(BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account)
-            + this.getVotingPowerForPeriod(BUTTERED_BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account);
+        uint256 lastCycleStart = lastClaimedBlockNumber - cycleLength;
+        uint256 multiplier = this.getTotalMultipliers(_account);
+        multiplier = multiplier == 0 ? 1e18 : multiplier;
+        uint256 breadVotingPower = this.getVotingPowerForPeriod(BREAD, lastCycleStart, lastClaimedBlockNumber, _account);
+        uint256 butteredBreadVotingPower =
+            this.getVotingPowerForPeriod(BUTTERED_BREAD, lastCycleStart, lastClaimedBlockNumber, _account);
+        return ((breadVotingPower + butteredBreadVotingPower) * multiplier) / PRECISION;
```
daopunk: I think instead of PRECISION, we should use WAD (wei as decimal) as that is more of a defi standard, small name, and can apply to any case of 1e18, rather than just in the case of precision.

src/YieldDistributor.sol
```diff
@@ -110,8 +111,13 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return this.getVotingPowerForPeriod(BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account)
-            + this.getVotingPowerForPeriod(BUTTERED_BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account);
+        uint256 lastCycleStart = lastClaimedBlockNumber - cycleLength;
+        uint256 multiplier = this.getTotalMultipliers(_account);
+        multiplier = multiplier == 0 ? 1e18 : multiplier;
+        uint256 breadVotingPower = this.getVotingPowerForPeriod(BREAD, lastCycleStart, lastClaimedBlockNumber, _account);
```
daopunk: instead of calling `this.contractFunction` on external functions, we can make the visibility public and then call directly

daopunk: so instead `this.getVotingPowerForPeriod` just `getVotingPowerForPeriod`

src/interfaces/IVotingMultipliers.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity ^0.8.22;
+
+import {IMultiplier} from "./multipliers/IMultiplier.sol";
+
+/// @title IVotingMultipliers
+/// @notice Interface for the VotingMultipliers contract
+/// @dev This interface defines the structure and functions for managing voting multipliers
+interface IVotingMultipliers {
+    /// @notice Thrown when attempting to add a multiplier that is already whitelisted
+    error MultiplierAlreadyWhitelisted();
```
daopunk: instead of "whitelist / blacklist", we should use "allowlist / denylist"

src/multipliers/NFTMultiplier.sol
```diff
@@ -0,0 +1,80 @@
...
+    /// @param user The address of the user to check
+    /// @return True if the user owns at least one NFT, false otherwise
+    function hasNFT(address user) public view override returns (bool) {
+        return nftContract.balanceOf(user) > 0;
+    }
+
+    /// @notice Get the multiplying factor for a given user
+    /// @param user The address of the user
+    /// @return The multiplying factor if the user owns an NFT, 0 otherwise
+    function getMultiplyingFactor(address user) external view override returns (uint256) {
```
daopunk: for internal variables, use underscore, so `user` should be `_user` to fit with other functions

src/multipliers/PermanentNFTMultiplier.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {INFTMultiplier} from "src/interfaces/multipliers/INFTMultiplier.sol";
+
+/// @title Permanent NFT Multiplier
+/// @notice This contract provides a permanent multiplying factor for users based on NFT ownership
+/// @dev Implements the INFTMultiplier interface
+contract PermanentNFTMultiplier is INFTMultiplier {
+    /// @notice The address of the NFT contract
+    IERC721 public immutable NFTAddress;
```
daopunk: immutable variables are formatted the same as constant; SNAKE_CASE

src/multipliers/PermanentNFTMultiplier.sol
```diff
@@ -0,0 +1,43 @@
...
+    /// @notice The multiplying factor applied to NFT holders
+    uint256 public immutable factor;
```
daopunk: should be FACTOR

src/test/MockMultiplier.sol
```diff
@@ -0,0 +1,31 @@
...
+/// @title MockMultiplier
+/// @notice A mock contract implementing the IMultiplier interface for testing purposes
+contract MockMultiplier is IMultiplier {
+    uint256 private _multiplyingFactor;
+    uint256 private _validUntil;
+
+    /// @notice Sets the multiplying factor and valid until block for testing
+    /// @param factor The multiplying factor to set
+    /// @param validUntilBlock The block number until which the multiplier is valid
+    function setMultiplier(uint256 factor, uint256 validUntilBlock) external {
```
daopunk: underscore for internal vars / function scoped vars / parameters; `_factor, _validUntilBlock`

test/YieldDistributor.t.sol
```diff
@@ -446,3 +450,295 @@ contract YieldDistributorTest is Test {
...
+    function testAddMultiplierRevertAlreadyWhitelisted() public {
+        yieldDistributor.addMultiplier(IMultiplier(address(mockMultiplier1)));
+        vm.expectRevert(IVotingMultipliers.MultiplierAlreadyWhitelisted.selector);
+        yieldDistributor.addMultiplier(IMultiplier(address(mockMultiplier1)));
+    }
+
+    function testRemoveMultiplier() public {
+        yieldDistributor.addMultiplier(IMultiplier(address(mockMultiplier1)));
+        yieldDistributor.removeMultiplier(IMultiplier(address(mockMultiplier1)));
+        vm.expectRevert();
```
daopunk, line 480: is there an error message to check for the revert?

test/YieldDistributor.t.sol
```diff
@@ -446,3 +450,295 @@ contract YieldDistributorTest is Test {
...
+        // Calculate expected voting power
+        uint256 expectedVotingPower = (initialVotingPower * (totalFactor)) / yieldDistributor.PRECISION();
+
+        // Check that the voting power was multiplied correctly
+        assertEq(yieldDistributor.projectDistributions(0), expectedVotingPower);
+        vm.stopPrank();
+    }
+
+    function testCastVote100Multipliers() public {
+        setUpVoteWithVariableMultipliers(100);
```
daopunk: nothing is asserted in these lasts tests

src/VotingMultipliers.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @notice Array of whitelisted multiplier contracts
+    IMultiplier[] public whitelistedMultipliers;
+
+    /// @notice Calculates the total multiplier for a given user
+    /// @param user The address of the user
+    /// @return The total multiplier value for the user
+    function getTotalMultipliers(address user) external view returns (uint256) {
+        uint256 totalMultiplier = 0;
```
daopunk: internal / function-scoped variables should have underscore; `_totalMultiplier`

src/VotingMultipliers.sol
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+        whitelistedMultipliers.push(_multiplier);
+        emit MultiplierAdded(_multiplier);
+    }
+
+    /// @notice Removes a multiplier from the whitelist
+    /// @param _multiplier The multiplier contract to be removed
+    function removeMultiplier(IMultiplier _multiplier) external onlyOwner {
+        bool isWhitelisted = false;
```
daopunk: make `isAllowListed` to be PC

src/YieldDistributor.sol
```diff
@@ -20,7 +21,7 @@ import {IYieldDistributor} from "src/interfaces/IYieldDistributor.sol";
  * @custom:coauthor kassandra.eth
  * @custom:coauthor theblockchainsocialist.eth
  */
-contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
+contract YieldDistributor is IYieldDistributor, OwnableUpgradeable, VotingMultipliers {
     /// @notice The address of the $BREAD token contract
     Bread public BREAD;
```
daopunk, line 27: add organization to contract vars;
```
    --- Registry ---
    
    /// @notice The address of the $BREAD token contract
    Bread public BREAD;
    
    --- Data ---
    
    /// @notice The minimum number of blocks between yield distributions
    uint256 public cycleLength;
    /// @notice The maximum number of points a voter can allocate to a project
    uint256 public maxPoints;
    ...
```

daopunk (changes requested):

src/YieldDistributor.sol
RonTuretzky: Resolved async

src/YieldDistributor.sol
daopunk: I think we should return a single number formatted to 1e18, to avoid confusion and enforce standardization

src/YieldDistributor.sol
daopunk: Is there a foreseeable problem deploying as a new proxy, thus changing the address? if there is not, i would prefer we deploy fresh, therefore do not need to worry about upgrade-safety, and not over-complicated things. but again, this is assuming there is no major problem with changing the address.

src/YieldDistributor.sol
RonTuretzky: addressed by 50787c1

RonTuretzky: addressed by 50787c1

src/interfaces/IVotingMultipliers.sol
RonTuretzky: addressed with 47333b3

src/VotingMultipliers.sol
daopunk: I feel like we are over-analyzing a non-issue. the gas costs are negligible here, and it's unlikely that we will be dealing with > 1000 voting-multipliers for the lifespan of this contract. thus i'm not in favor of spending time to highly optimize for gas-efficiency on a cheap L2

src/multipliers/NFTMultiplier.sol
RonTuretzky: addressed with 967d6cc

src/multipliers/PermanentNFTMultiplier.sol
RonTuretzky: addressed with 706cd8e

src/multipliers/PermanentNFTMultiplier.sol
RonTuretzky: addressed with 2d04f5c

src/test/MockMultiplier.sol
RonTuretzky: addressed with 6e78550

test/YieldDistributor.t.sol
RonTuretzky: resolved async

test/YieldDistributor.t.sol
RonTuretzky: addressed with f867e06

src/VotingMultipliers.sol
RonTuretzky: addressed with 80b7345

src/VotingMultipliers.sol
RonTuretzky: addressed with https://github.com/BreadchainCoop/breadchain/commit/47333b3a7ba237f0123d99b9298475ce09d865c0

src/YieldDistributor.sol
RonTuretzky: resolved async

src/VotingMultipliers.sol
RonTuretzky: @bagelface @rubydusa changes introduced in 0d4836e allow the frontend to query the multiplier indexes of the voter, and submit them along with the vote allocation. 

This reduces the complexity from O(n) to O(1) as the the contract will only address the multipliers specified. 

If the user specifies multipliers they do not have, the contract will simply award the 0 addition value (as the multiplier will return 0). Expiry is still checked. 

Assuming an honest frontend, this leads the the desired behavior with the gas overhead

daopunk (approved):

src/VotingMultipliers.sol
```diff
@@ -0,0 +1,104 @@
...
+/// @notice A contract for managing voting multipliers
+/// @dev Implements IVotingMultipliers interface
+contract VotingMultipliers is OwnableUpgradeable, IVotingMultipliers {
+    /// @notice Array of allowlisted multiplier contracts
+    IMultiplier[] public allowlistedMultipliers;
+
+    /// @notice Calculates the total multiplier for a given _user
+    /// @param _user The address of the _user
+    /// @return The total multiplier value for the _user
+    function getTotalMultipliers(address _user) public view returns (uint256) {
```
bagelface: Assuming we keeping this as a quality of life function so we can avoid the double query `getValidMultiplierIndexes(address)` -> `getTotalMultipliers(address, uint256[])`? If so, I think this should be moved to be closer to `getTotalMultipliers(address, uint256[])` so the overload is obvious. Maybe word adding a note in the natspec comments with a @dev tag

src/VotingMultipliers.sol
```diff
@@ -0,0 +1,104 @@
...
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+/// @title VotingMultipliers
+/// @notice A contract for managing voting multipliers
+/// @dev Implements IVotingMultipliers interface
+contract VotingMultipliers is OwnableUpgradeable, IVotingMultipliers {
+    /// @notice Array of allowlisted multiplier contracts
+    IMultiplier[] public allowlistedMultipliers;
+
+    /// @notice Calculates the total multiplier for a given _user
```
bagelface: Drop the underscore form `_user` when referring to it in the comments to be consistent with other comments.

src/interfaces/IVotingMultipliers.sol
```diff
@@ -0,0 +1,37 @@
...
+    /// @param multiplier The address of the added multiplier
+
+    event MultiplierAdded(IMultiplier indexed multiplier);
+    /// @notice Emitted when a multiplier is removed from the allowlist
+    /// @param multiplier The address of the removed multiplier
+    event MultiplierRemoved(IMultiplier indexed multiplier);
+    /// @notice Returns the multiplier at the specified index in the allowlist
+    /// @param index The index of the multiplier in the allowlist
+    /// @return The multiplier contract at the specified index
+
```
bagelface: remove line break

src/interfaces/IVotingMultipliers.sol
```diff
@@ -0,0 +1,37 @@
...
+interface IVotingMultipliers {
+    /// @notice Thrown when attempting to add a multiplier that is already allowlisted
+    error MultiplierAlreadyAllowlisted();
+    /// @notice Thrown when attempting to remove a multiplier that is not allowlisted
+    error MultiplierNotAllowlisted();
+    /// @notice Thrown when an invalid multiplier index is provided
+    error InvalidMultiplierIndex();
+    /// @notice Emitted when a new multiplier is added to the allowlist
+    /// @param multiplier The address of the added multiplier
+
```
bagelface: remove line break

src/interfaces/multipliers/ICrossChainProveableMultiplier.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {IProveableMultiplier} from "src/interfaces/multipliers/IProveableMultiplier.sol";
+
+/// @title Cross-Chain Proveable Multiplier Interface
+/// @notice Interface for contracts that provide a cross-chain proveable multiplying factor
+
```
bagelface: remove line break

src/interfaces/multipliers/IDynamicNFTMultiplier.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {INFTMultiplier} from "src/interfaces/multipliers/INFTMultiplier.sol";
+/// @title Dynamic NFT Multiplier Interface
+/// @notice Interface for contracts that provide a dynamic multiplying factor for _users based on NFT ownership
+/// @dev Extends the INFTMultiplier interface with dynamic multiplier functionality
+
```
bagelface: remove line break

src/interfaces/multipliers/IOnChainProveableMultiplier.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {IProveableMultiplier} from "src/interfaces/multipliers/IProveableMultiplier.sol";
+/// @title On-Chain Proveable Multiplier Interface
+/// @notice Interface for contracts that provide an on-chain proveable multiplying factor
+
```
bagelface: remove line break

src/interfaces/multipliers/IProveableMultiplier.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {IDynamicNFTMultiplier} from "src/interfaces/multipliers/IDynamicNFTMultiplier.sol";
+/// @title Proveable Multiplier Interface
+/// @notice Interface for contracts that provide a proveable multiplying factor based on _user activities
+
```
bagelface: remove line break

src/YieldDistributor.sol
bagelface: Deploying a new contract is definitely the move in this instance but I mean in future instances where we maybe want to make some changes to the `VotingMultipliers.sol` contract. Namespacing the storage layouts ensure `YieldDistributor` and `VotingMultipliers` are using their own storage locations that won't clash, so if we want to add a variable to one or the other, we can do so safely. I can make a quick branch so show what it would look like.

src/YieldDistributor.sol
```diff
@@ -227,6 +232,19 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
     }
 
+    /**
+     * @notice Cast votes for the distribution of $BREAD yield with multipliers
+     * @param _points List of points as integers for each project
+     * @param _multiplierIndices List of indices of multipliers to use for each project
+     */
+    function castVoteWithMultipliers(uint256[] calldata _points, uint256[] calldata _multiplierIndices) public {
+        uint256 _currentVotingPower = getCurrentVotingPower(msg.sender);
+        if (_currentVotingPower < minRequiredVotingPower) revert BelowMinRequiredVotingPower();
```
bagelface: Are we sure we want to check this requirement before multipliers are applied??

src/YieldDistributor.sol
```diff
@@ -110,17 +112,20 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return this.getVotingPowerForPeriod(BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account)
-            + this.getVotingPowerForPeriod(BUTTERED_BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account);
+        uint256 lastCycleStart = lastClaimedBlockNumber - cycleLength;
```
bagelface: Just realizing this will be invalid if `cycleLength` is changed. We need a "lastCycleLength" variable. Maybe it'd be good to create a `Cycle` struct?

src/YieldDistributor.sol
bagelface: I also think there is a strong case to be made for `VotingMultipliers` to be its own contract.

src/YieldDistributor.sol
RonTuretzky: trueeeee

src/YieldDistributor.sol
RonTuretzky: We should open a separate issue for this ... also mark it as bug so we don't upgrade changes without considering this...

src/VotingMultipliers.sol
RonTuretzky: addressed with a73cf97

src/YieldDistributor.sol
RonTuretzky: I think we all agree on this, resolving convo

src/YieldDistributor.sol
RonTuretzky: What are the pros/cons here? 

I believe we should redeploy fresh to avoid upgrade safety issues and address #96 at some point

src/VotingMultipliers.sol
RonTuretzky: This is more for frontend and for testing. Moved closer and added dev tag with feaeefd

src/interfaces/IVotingMultipliers.sol
RonTuretzky: addressed with 1d0e765

src/interfaces/IVotingMultipliers.sol
RonTuretzky: addressed with ba1a847

src/interfaces/multipliers/ICrossChainProveableMultiplier.sol
RonTuretzky: addressed with 2185940

src/interfaces/multipliers/IDynamicNFTMultiplier.sol
RonTuretzky: addressed with 610da88

src/interfaces/multipliers/IOnChainProveableMultiplier.sol
RonTuretzky: addressed with 47cd4e8

src/interfaces/multipliers/IProveableMultiplier.sol
RonTuretzky: addressed with 03d8935

src/YieldDistributor.sol
RonTuretzky: addressed with 586f8b0

src/YieldDistributor.sol
RonTuretzky: This is resolved by `previousCycleStartingBlock` , we no longer calculate the `lastCycleStart` variable, introduced this change from upstream (was introduced with #102)

RonTuretzky:
@daopunk @bagelface all issues have been resolved except the issue of inheritance vs modularity. 

As we plan to migrate to dough-kit sometime in the future, I would suggest we postpone dealing with this design question since we strongly agree that this should be modular and decoupled.

tldr: lets go with the path of least resistance for now as it will get reworked eventually anyways 

#110 has been opened to address this concern, but I suggest we hand that off to someone at a later point and unblock this pr which has been open since oct at this point

src/YieldDistributor.sol
RonTuretzky: #110 is also opened but I would really like to merge this and then maybe hand it off to someone else at a later point

bagelface (approved):

src/interfaces/multipliers/IDynamicNFTMultiplier.sol
```diff
@@ -0,0 +1,18 @@
...
+import {INFTMultiplier} from "src/interfaces/multipliers/INFTMultiplier.sol";
+/// @title Dynamic NFT Multiplier Interface
+/// @notice Interface for contracts that provide a dynamic multiplying factor for _users based on NFT ownership
+/// @dev Extends the INFTMultiplier interface with dynamic multiplier functionality
+interface IDynamicNFTMultiplier is INFTMultiplier {
+    /// @notice Get the multiplying factor for a _user
+    /// @param _user The address of the _user
+    /// @return The multiplying factor for the _user
+    function _userToFactor(address _user) external view returns (uint256);
```
daopunk, line 12: external and public functions should not have an underscore. internal functions should have underscores, and thus would not be part of the interface

src/multipliers/NFTMultiplier.sol
```diff
@@ -0,0 +1,80 @@
...
+
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {INFTMultiplier} from "src/interfaces/multipliers/INFTMultiplier.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+/// @title NFT Multiplier
+/// @notice Implementation of INFTMultiplier interface
+/// @dev Provides multiplying factors based on NFT ownership
+contract NFTMultiplier is INFTMultiplier, Initializable, OwnableUpgradeable {
```
daopunk, line 12: should we use Ownable2StepUpgradeable instead? it is safer to use during ownership transfers, but more gas

src/multipliers/PermanentNFTMultiplier.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
```
daopunk, line 4: specify import:
`import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";`

src/multipliers/PermanentNFTMultiplier.sol
```diff
@@ -0,0 +1,43 @@
...
+    /// @param _FACTOR The multiplying factor to apply to NFT holders
+    constructor(IERC721 _NFT_ADDRESS, uint256 _FACTOR) {
+        NFT_ADDRESS = _NFT_ADDRESS;
+        FACTOR = _FACTOR;
+    }
+
+    /// @notice Get the multiplying factor for a _user
+    /// @param _user The address of the _user
+    /// @return The multiplying factor if the _user has an NFT, 0 otherwise
+    function getMultiplyingFactor(address _user) external view override returns (uint256) {
```
daopunk, line 27: missing declared return value. 
`_factor = hasNFT(_user) ? FACTOR : 0;`
instead of using return keyword. this applies throughout the repo

daopunk (changes requested):

src/multipliers/PermanentNFTMultiplier.sol
RonTuretzky: addressed in #123

src/multipliers/PermanentNFTMultiplier.sol
RonTuretzky: resolved by #124

src/multipliers/NFTMultiplier.sol
RonTuretzky: resolved by #124

src/interfaces/multipliers/IDynamicNFTMultiplier.sol
RonTuretzky: resolved by #124

Issue #84, opened 2024-08-29
getVotingPowerForPeriod optimization


RonTuretzky:
Iterate on suggestion opened in #65 and [here](https://github.com/BreadchainCoop/breadchain/pull/65#discussion_r1728647354)

PR #101, opened 2024-09-28
Reducing complexity of voting power function

RonTuretzky:
#84

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -142,8 +142,8 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
 
         uint256 _totalVotingPower;
 
-        for (uint32 i = _numCheckpoints; i > 0;) {
-            _currentCheckpoint = _sourceContract.checkpoints(_account, --i);
+        for (uint32 i = _numCheckpoints - 1; --i > 0;) {
+            _currentCheckpoint = _sourceContract.checkpoints(_account, i);
 
             if (_currentCheckpoint._key <= _end) {
                 uint48 _effectiveStart = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -142,8 +142,8 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
 
         uint256 _totalVotingPower;
 
-        for (uint32 i = _numCheckpoints - 1; --i > 0;) {
-            _currentCheckpoint = _sourceContract.checkpoints(_account, i);
+        for (uint32 i = _numCheckpoints; i > 0;) {
+            _currentCheckpoint = _sourceContract.checkpoints(_account, --i);
 
             if (_currentCheckpoint._key <= _end) {
                 uint48 _effectiveStart = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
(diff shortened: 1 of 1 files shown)


src/YieldDistributor.sol
```diff
@@ -140,32 +140,18 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
-        } while (_currentCheckpoint._key > _end);
-
-        /// Initialize voting power with the latest checkpoint thats within the interval (or nearest to it)
-        uint48 _latestKey = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
-        uint256 _totalVotingPower = _currentCheckpoint._value * (_end - _latestKey);
-
-        if (_latestKey == _start) return _totalVotingPower;
+        uint256 _totalVotingPower;
 
         for (uint32 i = _currentCheckpointIndex; i > 0;) {
```
bagelface: with this refactor moving the logic into a single for loop, I think the variable `_currentCheckpointIndex` would be better renamed to `_numCheckpoints` and then when setting us this for loop you can instead do:

```
for (uint32 i = _numCheckpoints - 1; i >= 0; --i)` {
  _currentCheckpoint = _sourceContract.checkpoints(_account, i);
  ...
}
```

src/YieldDistributor.sol
```diff
@@ -140,32 +140,18 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
-            /// Latest checkpoint voting power is calculated when initializing `_totalVotingPower`, so we pre-decrement the index here
             _currentCheckpoint = _sourceContract.checkpoints(_account, --i);
+            if (_currentCheckpoint._key <= _end) {
+                uint48 effectiveStart = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
```
bagelface: prefix `_` for internal variable

src/YieldDistributor.sol
```diff
@@ -140,32 +140,18 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
-        uint48 _latestKey = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
-        uint256 _totalVotingPower = _currentCheckpoint._value * (_end - _latestKey);
-
-        if (_latestKey == _start) return _totalVotingPower;
+        uint256 _totalVotingPower;
 
         for (uint32 i = _currentCheckpointIndex; i > 0;) {
-            /// Latest checkpoint voting power is calculated when initializing `_totalVotingPower`, so we pre-decrement the index here
             _currentCheckpoint = _sourceContract.checkpoints(_account, --i);
+            if (_currentCheckpoint._key <= _end) {
```
bagelface: line break before big if block for readability

src/YieldDistributor.sol
```diff
@@ -140,32 +140,18 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
...
+                uint48 effectiveStart = _currentCheckpoint._key < _start ? uint48(_start) : _currentCheckpoint._key;
+                _totalVotingPower += _currentCheckpoint._value * (_end - effectiveStart);
 
-            /// Add voting power for the sub-interval to the total
-            _totalVotingPower += _currentCheckpoint._value * (_latestKey - _currentCheckpoint._key);
+                if (effectiveStart == _start) break;
```
bagelface, line 151: can prob remove this line break and combine these two since they're both single line control flow statements

bagelface (approved):
a few suggested changes, but logic changes seem fine

src/YieldDistributor.sol
RonTuretzky: addressed with 9f92e21

src/YieldDistributor.sol
RonTuretzky: addressed with b6c1b77

src/YieldDistributor.sol
RonTuretzky: addressed with 51931c6

src/YieldDistributor.sol
RonTuretzky: addressed with d87945a

ShaNagbaImuru (approved):
LGTM, thanks!

PR #102, opened 2024-10-11
Derive cycle length from actual distribution blocks 

### LINKED ISSUE #82: Derive cycle length from actual blocks of distribution
As explained [here](https://github.com/BreadchainCoop/breadchain/pull/65/files/a42230e432ab989eeeff47883b10c9037dd9c86b#r1676297379) 

- [x] When a distribution occurs, the distribution interval should be updated to be the current block and the last distribution block 
- [x] Voting power should be reconfigured to be calculated from the last actual interval between distributions

RonTuretzky:
#82
Upgrade safety validation as performed

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -52,7 +52,7 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     /// @notice The address of the `ButteredBread` token contract
     ERC20VotesUpgradeable public BUTTERED_BREAD;
     /// @notice The block number before the last yield distribution
-    uint256 public beforeLastClaimedBlockNumber;
+    uint256 public previousCycleStartingBlock;
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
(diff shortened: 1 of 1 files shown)


src/YieldDistributor.sol
```diff
@@ -51,6 +51,8 @@ contract YieldDistributor is IYieldDistributor, OwnableUpgradeable {
     uint256 public yieldFixedSplitDivisor;
     /// @notice The address of the `ButteredBread` token contract
     ERC20VotesUpgradeable public BUTTERED_BREAD;
+    /// @notice The block number before the last yield distribution
+    uint256 public beforeLastClaimedBlockNumber;
```
bagelface: maybe `previousCycleStartingBlock` would be easier to understand? like I get what it is, but the name is a bit confusing to me

bagelface (approved):
lgtm just one small comment

src/YieldDistributor.sol
RonTuretzky: good suggestion!

RonTuretzky: addressed with 6a814e4

PR #104, opened 2024-10-17
 QoL view functions for frontend 

RonTuretzky:


src/ButteredBread.sol
```diff
@@ -123,6 +123,14 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
         revert NonDelegatable();
     }
 
+    /// @notice Get the balance of a specific LP for a given account
+    /// @param holder The address of the account to get the balance for
+    /// @param lp The address of the LP to get the balance for
+    /// @return LPData memory The balance of the LP for the given account
+    function balanceOfLP(address holder, address lp) external view returns (LPData memory) {
```
bagelface: use argument names `_holder` and `_lp` to be consistent with other functions. make sure to change them in the natspec comments too pls 🙏

bagelface (approved):
small comment then lgtm

src/ButteredBread.sol
RonTuretzky: addressed with 56958fd :)

Issue #116, opened 2024-12-18
perform `IERC20(_lp).transferFrom` in the end of `_deposit` in `ButteredBread`


lirona:
to adhere to the checks-effects-interactions pattern

Issue #119, opened 2024-12-18
`deposit` changes the user's scaling factor without syncing voting weight in `ButteredBread`


lirona:
in case a user performs a second deposit and the scaling factor has changed since the last deposit, the user's amount of buttered bread received against the first deposit will not be adjusted to fit the new scaling factor (unless the user's address was passed in the array of users in `modifyScalingFactor`, which could not happen as mentioned [in this comment](https://github.com/BreadchainCoop/breadchain/blob/01af940e6a6231b8945995c55176f1febf3e1b2b/src/ButteredBread.sol#L153))

if this is not done, then even though `_syncVotingWeight` is done on `withdraw` to ensure that the user is withdrawing the updated amount of lp tokens, the user's buttered bread balance would not be updated to the right amount and the withdraw would try to burn the wrong amount of buttered bread from the user, and the withdraw would fail. `_syncVotingWeight` should also be done on `deposit` to ensure that the voting power stays updated.

example:
- scaling factor is 100.
- user deposits 100 lp tokens and receives 100 buttered bread.
- scaling factor changes to 150 but user does not get updated.
- user deposits another 100 lp tokens, her scaling factor changes to 150 and receives 150 buttered bread.
- user withdraws her 200 lp tokens, `_syncVotingWeight` does nothing since the user's voting weight is already updated, the function tries to burn 200*150/100 = 300 buttered bread from the user, but the user only has 250 buttered bread => function reverts with `ERC20InsufficientBalance(user, 250, 300)`

if fixed:
- scaling factor is 100.
- user deposits 100 lp tokens and receives 100 buttered bread.
- scaling factor changes to 150 but user does not get updated.
- user deposits another 100 lp tokens, `_syncVotingWeight` updates her scaling factor to 150 and mints her an extra 50 buttered bread, and then receives 150 buttered bread.
- user withdraws her 200 lp tokens, `_syncVotingWeight` does nothing since the user's voting weight is already updated, the function burns 200*150/100 = 300 buttered bread from the user (which she has), and user receives 200 lp tokens 🤗

lirona:
this can also be fixed by keeping an array of all users per lp and iterating on it when changing scaling factor to ensure no one gets left out, and that would allow to remove the syncing of voting weight from `deposit` and `withdraw` and take that gas cost off of the users. that will not be possible in one transaction if user base gets big, but the syncing could be broken down into multiple transactions.

RonTuretzky:
[...]

Issue #117, opened 2024-12-18
`ButteredBread` contract assumes lp tokens always has 18 decimals


lirona:
in the case that it doesn't, the amount of buttered bread received will not be the same as the amount of lp tokens deposited. the contract could be generalized and do the conversion between the decimals on `deposit` and `withdraw` or check when adding a new lp that it has 18 decimals. both these solutions would requiring copying `IERC20` to your codebase and adding `decimals()`, since it's not part of the [OZ implementation of the interface](https://github.com/openzeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol).

to the very least there should be a warning on the code so that this issue is not overlooked in case of adding more lps in the future.

Issue #118, opened 2024-12-18
function `balanceOfLP` returns `LPData` and not balance in `ButteredBread`


lirona:
the function name and natspec say it returns the balance but it returns `LPData` which contains balance and scaling factor.
there is another getter called `accountToLPBalance` which returns the balance.
consider removing this function if only the balance is needed, or changing it's name. for now i'll change the natspec at least :)

RonTuretzky:
@secbajor @subject026  do you know the `balanceOfLP` is used in the crowdstaking app?

subject026:
Not right now

RonTuretzky:
> Not right now

Is it planned in any way? Otherwise we might rename it

subject026:
We will need to be able to fetch the scaling factors although I think being able to fetch them on their own rather than paired with a LP balance would be more flexible

RonTuretzky:
> We will need to be able to fetch the scaling factors although I think being able to fetch them on their own rather than paired with a LP balance would be more flexible

Yea we have another method that exposes it per lp. Great so we can remove this method I think @lirona ! Unless it's used in tests somewhere 🤔

lirona:
hmm i don't see a getter for scaling factors per user anywhere.. @subject026 did you mean you need them per user and lp or just per lp?

RonTuretzky:
I believe there's no reason we would need the modifier on the frontend at all, as there's another method that exposes the voting power and the modifier is accounted for in there

PR #120, opened 2024-12-19
liron's `ButteredBread` fixes

lirona:
for #119 i've implemented the simple solution but you could also just make sure you've iterated over all users and that way you don't need to pass an array of users to update and don't need to keep scaling factor per user in the contract or to sync voting weights.

for #117 i've just added a warning, as decided by @RonTuretzky

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -135,14 +135,14 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
 
     /// @notice Deposit LP tokens and mint ButteredBread with corresponding LP scaling factor
     function _deposit(address _account, address _lp, uint256 _amount) internal {
-        require(IERC20(_lp).transferFrom(_account, address(this), _amount), "Transfer failed");
+        _syncDelegation(_account);
         _accountToLPData[_account][_lp].balance += _amount;
 
         uint256 currentScalingFactor = scalingFactors[_lp];
         _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
 
         _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
-        _syncDelegation(_account);
+        require(IERC20(_lp).transferFrom(_account, address(this), _amount), "Transfer failed");
 
         emit ButterAdded(_account, _lp, _amount);
     }
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -5,6 +5,8 @@ import {OwnableUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/a
 import {ERC20VotesUpgradeable} from
     "openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+import {ReentrancyGuardUpgradeable} from
+    "openzeppelin-contracts-upgradeable/contracts/utils/ReentrancyGuardUpgradeable.sol";
 
 import {IButteredBread} from "src/interfaces/IButteredBread.sol";
 import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
@@ -17,7 +19,7 @@ import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
  * @custom:coauthor @daopunk
  * @custom:coauthor @bagelface
  */
-contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradeable {
+contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradeable, ReentrancyGuardUpgradeable {
     /// @notice Value used for calculating the precision of scaling factors
     uint256 public constant FIXED_POINT_PERCENT = 100;
     /// @notice `IERC20Votes` contract used for powering `ButteredBread` voting
@@ -47,6 +49,8 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
 
         __Ownable_init(msg.sender);
         __ERC20_init(_initData.name, _initData.symbol);
+        __ERC20Votes_init();
+        __ReentrancyGuard_init();
 
         for (uint256 i; i < _initData.liquidityPools.length; ++i) {
             scalingFactors[_initData.liquidityPools[i]] = _initData.scalingFactors[i];
@@ -75,7 +79,7 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
      * @param _lp Liquidity Pool token
      * @param _amount Value of LP token
      */
-    function deposit(address _lp, uint256 _amount) external onlyAllowed(_lp) {
+    function deposit(address _lp, uint256 _amount) external onlyAllowed(_lp) nonReentrant {
         if (_amount == 0) revert AmountZero();
         _deposit(msg.sender, _lp, _amount);
     }
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -125,10 +125,10 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
         revert NonDelegatable();
     }
 
-    /// @notice Get the balance of a specific LP for a given account
-    /// @param _holder The address of the account to get the balance for
-    /// @param _lp The address of the LP to get the balance for
-    /// @return LPData memory The balance of the LP for the given account
+    /// @notice Get the balance and scaling factor of a specific LP for a given account
+    /// @param _holder The address of the account to get the data for
+    /// @param _lp The address of the LP to get the data for
+    /// @return LPData memory The balance and scaling factor of the LP for the given account
     function balanceOfLP(address _holder, address _lp) external view returns (LPData memory) {
         return _accountToLPData[_holder][_lp];
     }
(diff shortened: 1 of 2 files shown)


src/ButteredBread.sol
```diff
@@ -75,6 +75,7 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
      * @param _amount Value of LP token
      */
     function deposit(address _lp, uint256 _amount) external onlyAllowed(_lp) {
+        require(_amount != 0, "Amount must be greater than 0");
```
bagelface: prefer to use "custom errors" approach for assertions.

src/ButteredBread.sol
```diff
@@ -84,12 +85,14 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
      * @param _amount Value of LP token
      */
     function withdraw(address _lp, uint256 _amount) external onlyAllowed(_lp) {
+        require(_amount != 0, "Amount must be greater than 0");
```
bagelface: same here comment about "custom errors" here

src/ButteredBread.sol
```diff
@@ -155,7 +158,7 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
         _accountToLPData[_account][_lp].balance -= _amount;
 
         _burn(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
-        IERC20(_lp).transfer(_account, _amount);
+        require(IERC20(_lp).transfer(_account, _amount), "Transfer failed");
```
bagelface: prefer custom errors and something like:
```
bool success = IERC20(_lp).transfer(_account, _amount);
if (!success) revert TransferFailed();

```

src/ButteredBread.sol
```diff
@@ -123,25 +126,25 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
...
-
-        _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
         _syncDelegation(_account);
 
+        /// @dev ensure proper accounting in case of admin error in `modifyScalingFactor` where not all holders are updated
+        _syncVotingWeight(_account, _lp);
+        _accountToLPData[_account][_lp].balance += _amount;
+        
+        _mint(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
+        require(IERC20(_lp).transferFrom(_account, address(this), _amount), "Transfer failed");
```
bagelface: custom error pattern like:
```
bool success = IERC20(_lp).transferFrom(_account, address(this), _amount)
if (!success) revert TransferFailed();
```
for better readability

src/ButteredBread.sol
```diff
@@ -123,25 +126,25 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
...
-        IERC20(_lp).transferFrom(_account, address(this), _amount);
-        _accountToLPData[_account][_lp].balance += _amount;
-
-        uint256 currentScalingFactor = scalingFactors[_lp];
-        _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
-
-        _mint(_account, _amount * currentScalingFactor / FIXED_POINT_PERCENT);
         _syncDelegation(_account);
 
+        /// @dev ensure proper accounting in case of admin error in `modifyScalingFactor` where not all holders are updated
```
bagelface, line 150: i think the comment isn't necessary here. it can potentially be moved to the natspec function comment

bagelface (changes requested):
a few minor changes but overall this these are good updates

src/ButteredBread.sol
lirona: that was the way it was done by you in the `withdraw` function.. do you want me to move that from there as well?

lirona:
added custom errors, note that if you [upgrade your solidity version](https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement) you can use the new and much nicer `require(bool, error)` pattern :)

and please take note of the issues i mentioned, issue #117 is high/critical severity, and issue #119 is critical severity.. but that's fine if you're happy with the solutions in this pr

src/ButteredBread.sol
```diff
@@ -123,25 +126,27 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, OwnableUpgradea
...
+        _syncVotingWeight(_account, _lp);
+        _accountToLPData[_account][_lp].balance += _amount;
+        
+        _mint(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
+        
+        bool success = IERC20(_lp).transferFrom(_account, address(this), _amount);
```
daopunk: Why is transfer moved after mint? Funds should arrive to this contract, then tokens minted to the depositor.

daopunk (changes requested):

src/ButteredBread.sol
lirona: to follow the checks-effects-interactions pattern.
it's in the [commit message](https://github.com/BreadchainCoop/breadchain/pull/120/commits/c210616e12eeab396cb9ed15d3ab17cab125ab8c) :)

src/ButteredBread.sol
bagelface: oh i didnt notice. Yea up to you if you want to change it in withdraw, I'll disregard this in the review.

bagelface (approved):
lgtm

src/ButteredBread.sol
lirona: actually would be good to add a comment in the public functions' natspec, i added it in [this commit](https://github.com/BreadchainCoop/breadchain/pull/120/commits/4e801486fdd8b28ae982455fdef20d9a0b815436), wdyt?

src/ButteredBread.sol
lirona: to follow the checks-effects-interactions pattern, which reduces potential vulnerabilities like reentrancy.
and anyway it all happens in the same transactions so it either all happens or it all doesn't happen.. the order doesn't change the outcome.
actually to make the review process easier, i submitted issues for each fix and linked them to each commit that fixes them. this one fixes #116
welcome to look at the rest :)
#114 #115 #117 #118 #119

RonTuretzky: Here is a nice blog posts on this [pattern](https://samvishwas.medium.com/solidity-programming-the-checks-effects-interactions-pattern-bac0e5a5f659#:~:text=The%20first%20step%20in%20the,proceed%20with%20the%20requested%20action.)

RonTuretzky (approved):
LGTM!

src/ButteredBread.sol
daopunk: @RonTuretzky i wrote a article on this too few years ago 
https://medium.com/better-programming/solidity-smart-contract-security-preventing-reentrancy-attacks-fc729339a3ff

daopunk: I get it is to follow the CEI pattern, but that usually applies to funds flowing out to the caller. in this case, funds are flowing into the contract, only then will coins be minted to the caller. so in this context, it seems minting first before receiving payment is more of a vulnerability, right?

RonTuretzky: If I'm understanding correctly, what we're afraid of is a re-entrancy attack from the LP. 

If the re-entrancy entrypoint is in the `transferFrom` , it means the attack uses `transferFrom` to call `deposit` again

so this is the attack in the currently deployed implementation

```mermaid 
sequenceDiagram 
autonumber
Attacker ->> BB : deposit
loop  
BB ->> CorruptedLP : transferFrom
CorruptedLP ->> BB : deposit
end
BB ->> BB : mint
```

I believe the proposed change would make it so that this is possible instead 

```mermaid 
sequenceDiagram 
autonumber
Attacker ->> BB : deposit
loop  
BB ->> BB : mint
BB ->> CorruptedLP : transferFrom
CorruptedLP ->> BB : deposit
end
```

Which seems worst to me

daopunk: Exactly.

Notice in the _withdraw function, the balance is updated before being sent to the caller, whereas in the _deposit function the balance is updated after the funds are received to the contract. It is the same pattern but applied in reverse depending on the context.

receive funds from caller => mint coins to caller
burn coins from caller => release funds to caller

RonTuretzky: updated the diagram to reflect the mint before transferFrom

daopunk: Might be worth it to add a re-entrancy guard tho @lirona

lirona: great catch! sorry guys, didn't think this through.. good lesson for me that best practices are not always best :))
i will revert this change and add a reentrancy guard 🙏

lirona: moved back in https://github.com/BreadchainCoop/breadchain/pull/120/commits/4fc9568be5c3ac64a0e28e082bf3ebf407e27761 and added reentrancy guard in https://github.com/BreadchainCoop/breadchain/pull/120/commits/73bc7fc447979812b2856913bea23733ef0a755b
didn't add on the `withdraw` function yet, on one hand can't hurt, but it will add some gas cost (to reduce the gas cost we can implement it using [`transient` which is supposed to reduce the gas cost significantly](https://docs.soliditylang.org/en/v0.8.28/contracts.html#transient-storage), but i'm guessing OZ will update their implementation to it soon)
wdyt?

RonTuretzky: I guess we should add it to the withdraw flow as well, as there could be a re-entrancy as well over there if we're already covering it

lirona: added, so resolving this :)

daopunk (approved):

PR #124, opened 2025-01-04
123 voting multiplier hotfix

### LINKED ISSUE #123: Voting Multiplier Hotfix
- [x]  [external and public functions should not have an underscore. internal functions should have underscores, and thus would not be part of the interface](https://github.com/BreadchainCoop/breadchain/pull/100/#discussion_r1901196853)
- [x] [specify import:
import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";](https://github.com/BreadchainCoop/breadchain/pull/100/#discussion_r1901198370)
- [x] [should we use Ownable2StepUpgradeable instead? it is safer to use during ownership transfers, but more gas](https://github.com/BreadchainCoop/breadchain/pull/100/#discussion_r1901197752)
- [ ] [missing declared return value.
_factor = hasNFT(_user) ? FACTOR : 0;
instead of using return keyword. this applies throughout the repo](https://github.com/BreadchainCoop/breadchain/pull/100/#discussion_r1901199655)

RonTuretzky:
#123 Addresses this issue, from final comments from @daopunk on #100 

I opened a temporary branch [here](https://github.com/BreadchainCoop/breadchain/tree/stash_return_value_rewrite) with some progress on the final task, but I'm unsure whether it's worth addressing this final task as this repo is not in the wonderland template anyways

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity 0.8.25;
 
-import {OwnableUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
 import {ERC20VotesUpgradeable} from
     "openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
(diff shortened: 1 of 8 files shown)


bagelface (approved):
lgtm, although it doesn't address the final bullet of #123

daopunk (approved):

Issue #127, opened 2025-03-03
Implement VotingBoostMultiplier


secbajor:
Background
We have scoped the voting boost feature on the front end so that in it's first iteration we want to display and make available a single voting boost which should be the Voting Streak boost. The Voting Streak boost should apply a voting multiplier if the account has voted in X past subsequent voting cycles. We need to write the smart contract to enable this voting boost. 

Actions
1.  Add a new contract like VotingBoostMultiplier.sol to src/multipliers/ 
2. Implement the getMultiplyingFactor function -- it should check the user’s voting history (from yield distributor contract) to return the multiplier. 
2a. With a streak of 2 voting cycles, the boost should be x1.02
2b. 4 voting cycles, x1.04
2c. 6+ voting cycles, x1.06
3. Implement validUntil function -- this may be achievable by getting the cycleLength from the YieldDistributor, since this boost should basically be valid until the end of the next cycle for as long as the account has voted in the previous cycle

RonTuretzky:
@secbajor I believe some work was already done here https://github.com/BreadchainCoop/breadchain/commits/113-voting-streak-multiplier/

secbajor:
@RonTuretzky OK i just looked into that, the branch is way out of date with dev so i found it hard to go over. 

I opened [this PR](https://github.com/BreadchainCoop/breadchain/pull/128/files) that I think adds the minimal changes from -113 that i intended to be in scope with this issue. @bagelface since we were working together on this previously can you take a look at my questions on the PR?

PR #128, opened 2025-03-19
Add Voting streak multiplier contract

secbajor:
[#127]

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -10,7 +10,6 @@ import {Bread} from "bread-token/src/Bread.sol";
 
 import {IYieldDistributor} from "src/interfaces/IYieldDistributor.sol";
 import {VotingMultipliers} from "src/VotingMultipliers.sol";
-import {IVotingStreakMultiplier} from "src/interfaces/multipliers/IVotingStreakMultiplier.sol";
 
 /**
  * @title Breadchain Yield Distributor
@@ -56,19 +55,21 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
     ERC20VotesUpgradeable public BUTTERED_BREAD;
     /// @notice The block number before the last yield distribution
     uint256 public previousCycleStartingBlock;
-    /// @notice Mapping of user addresses to their current multiplier factor
-    mapping(address => uint256) public userToMultiplier;
-    /// @notice Mapping of user addresses to their multiplier validity period
-    mapping(address => uint256) public userToValidity;
-    /// @notice The maximum multiplier incrementation
-    uint256 public maxVotingStreakMultiplier;
-    /// @notice The increment value for the multiplier
-    uint256 public votingStreakMultiplierIncrement;
+
+    struct Cycle {
+        uint256 startBlock; // block number cycle started at
+        uint256 endBlock; // block number cycle ended at
+        uint256 totalVotes; // total number of votes
+        address[] projects; // list of projects
+        uint256[] voteDistribution; // distribution of votes
+        mapping(address => bool) voted; // addresses that voted
+    }
+
+    mapping(uint256 => Cycle) public cycles; // Mapping of cycle ID to Cycle struct
 
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor(address _votingStreakMultiplier) {
+    constructor() {
         _disableInitializers();
-        votingStreakMultiplier = IVotingStreakMultiplier(_votingStreakMultiplier);
     }
 
     function initialize(
@@ -206,6 +207,9 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         (bool _resolved,) = resolveYieldDistribution();
         if (!_resolved) revert YieldNotResolved();
 
+        cycles[previousCycleStartingBlock].voteDistribution = projectDistributions;
+        cycles[previousCycleStartingBlock].totalVotes = currentVotes;
+
         BREAD.claimYield(BREAD.yieldAccrued(), address(this));
         previousCycleStartingBlock = lastClaimedBlockNumber;
         lastClaimedBlockNumber = block.number;
@@ -213,6 +217,13 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         uint256 _fixedYield = balance / yieldFixedSplitDivisor;
         uint256 _baseSplit = _fixedYield / projects.length;
         uint256 _votedYield = balance - _fixedYield;
+
+        // start a new cycle
+        cycles[lastClaimedBlockNumber] = Cycle({
+            startBlock: lastClaimedBlockNumber,
+            endBlock: lastClaimedBlockNumber + cycleLength,
+            projects: projects
+        });
 
         for (uint256 i; i < projects.length; ++i) {
             uint256 _votedSplit = ((projectDistributions[i] * _votedYield * PRECISION) / currentVotes) / PRECISION;
@@ -239,17 +250,6 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         _castVote(msg.sender, _points, _currentVotingPower);
     }
 
-    /// @notice Calculate and return the voting streak multiplier for a user
-    /// @param _account Address of the user to calculate the multiplier for
-    /// @return uint256 The calculated voting streak multiplier
-    function votingStreakMultiplier(address _account) public view returns (uint256) {
-        uint256 validity = userToValidity[_account];
-        if (block.number > validity) {
-            return 0; // No active multiplier
-        }
-        return userToMultiplier[_account]; // Return the current multiplier
-    }
-
     /**
      * @notice Cast votes for the distribution of $BREAD yield with multipliers
      * @param _points List of points as integers for each project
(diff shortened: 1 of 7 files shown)


src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,137 @@
...
+import {ERC721Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import {IDynamicNFTMultiplier} from "src/interfaces/multipliers/IDynamicNFTMultiplier.sol";
+import {YieldDistributor} from "src/YieldDistributor.sol";
+import {IVotingStreakMultiplier} from "src/interfaces/multipliers/IVotingStreakMultiplier.sol";
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+/// @title VotingStreakMultiplier
+/// @notice A contract for managing voting streak multipliers as NFTs
```
secbajor, line 12: can i get some context why the idea here is to manage voting streak multiplier as an NFT? i am missing a lot of background knowledge, so looking for just a high level answer i can use as a reference while i dig into this.

if i understand correctly the way its intended to work is something like 
- a new cycle begins, and an nft is generated by this contract
- when a wallet votes, they are minted the nft for that cycle 
- now, the multiplier can be granted for wallets that own the nft

secbajor (commented):
Could this be mergeable as-is or would there be other changes recommended?

src/multipliers/VotingStreakMultiplier.sol
bagelface: I don't think the voting streak multiplier should be an NFT multiplier since it's a multiplier that is validated based on state that **should** be directly queryable from the YieldDistributor contract. I think you should be able to make it work just using the [IMultiplier.sol](https://github.com/BreadchainCoop/breadchain/blob/dev/src/interfaces/multipliers/IMultiplier.sol) interface.

For `validUntil` I think you can just return the max value for a uint256 because the YieldDistributor contract is eternal, so a voting streak multiplier would also have an eternal life span.

For `getMultiplyingFactor` I think we might actually be missing some data for this in the current implementation of "YieldDistributor". Right now, we're only storing `accountLastVoted` and `voterDistributions` data. If we're looking back multiple cycles, we'll need to add something like `mapping(address => mapping(uint256 => bool)) accountCyclesVoted` or maybe even a `struct Cycle` that includes a member like `mapping(address => bool) addressesVoted` so we can see exactly who participated in a given cycle.

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -30,10 +30,10 @@ contract VotingStreakMultiplier is
     YieldDistributor public yieldDistributor;
 
     /// @notice Mapping of user addresses to their current multiplier factor
-    mapping(address => uint256) public override userToFactor;
+    mapping(address => uint256) public userToMultiplier;
 
     /// @notice Mapping of user addresses to their multiplier validity period
-    mapping(address => uint256) public override userToValidity;
+    mapping(address => uint256) public userToValidity;
```
secbajor: probably add these to IMultiplier and override them here

bagelface: Not sure we need to add them to the `IMultiplier` interface. It's unclear to me how they'd be used differently than the existing functions that are `f(address) returns uint256` for multliplying factor and validity

src/YieldDistributor.sol
```diff
@@ -43,22 +44,31 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
     /// @notice The block number before the last yield distribution
     uint256 public previousCycleStartingBlock;
+    /// @notice Mapping of user addresses to their current multiplier factor
+    mapping(address => uint256) public userToMultiplier;
+    /// @notice Mapping of user addresses to their multiplier validity period
+    mapping(address => uint256) public userToValidity;
+    /// @notice The maximum multiplier incrementation
+    uint256 public maxVotingStreakMultiplier;
+    /// @notice The increment value for the multiplier
+    uint256 public votingStreakMultiplierIncrement;
```
RonTuretzky: confused on why this is in the yield distributor, this should be implemented in the multiplier no?

src/YieldDistributor.sol
```diff
@@ -43,22 +44,31 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
 
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() {
+    constructor(address _votingStreakMultiplier) {
```
RonTuretzky: we should need to add them through a setter, not through the constructor 

also because these contracts are upgradeable, we actually don't use the constructor at all 

https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable

src/YieldDistributor.sol
secbajor: this commit was exploring a design where the yielddistributor itself knows about the voting history of addresses.

i have some feedback from bagel to integrate that will change this significantly, so ignore this for now :)

src/YieldDistributor.sol
```diff
@@ -200,30 +204,38 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+        cycles[currentCycle].voteDistribution = projectDistributions;
+        cycles[currentCycle].totalVotes = currentVotes;
+        cycles[currentCycle].endBlock = lastClaimedBlockNumber;
+        currentCycle++;
+    }
+
+    /// @notice Initialize the new cycle
+    function initializeCycle() internal {
+        cycles[currentCycle].startBlock = lastClaimedBlockNumber;
+        cycles[currentCycle].projects = projects;
```
RonTuretzky, line 218: why aren't we tracking the cycles in the multiplier itself per user instead of adding additional logic to the yd? I'm against this approach

RonTuretzky (changes requested):

src/YieldDistributor.sol
bagelface: Can you elaborate on the solution you're proposing? Tracking the cycles in the multiplier would require updating storage, so it's not immediately clear to me how that would work with the existing implementation of the multipliers extension where all the functions are "view".

With this solution, yes, we modify the YieldDistributor contract, but we add only a single storage write or two (depending on implementation) with each cycle start/end and add additional data that can be useful to implementing other multipliers.

To be clear, I'm not necessarily endorsing this specific implementation, but this approach overall (modifying the YD vs tracking this data in the multiplier)

RonTuretzky: This breaks the entire pattern and design of the multiplier spec, which is meant to not add multiplier logic to the YD but instead provide a way to add modular and decoupled voting power logic... 

if we want this data on the YD  in general that's another question, but as far as I can tell of those data structures can sit inside the multiplier and I can't think of other multipliers that would use this. 

I'm very against adding any code to the YD in general,and I don't understand why we need all this data for the voting streak multiplier. Don't we just need a counter that goes up when the user votes? 

My suggestion would be to hook into the multiplier when the user votes , update their last voted time and increment a counter. Next time they vote, we can use the CYCLE_LENGTH to determine whether they've broken their streak or not by referencing whether their last time voted is longer than CYCLE_LENGTH

bagelface: "My suggestion would be to hook into the multiplier when the user votes"
What exactly do you mean by this? How can we "hook" into the multiplier when the user votes without updating the YieldDistributor or adding an additional transaction? I guess we can add a contract that wraps the vote transaction and updates multipliers in the same transaction, but that just seems awkward. Feel free to DM me to discuss.

fwiw, this is not a suggestion to "add multiplier logic to the YD". It is a suggestion to add storage variables for data that didn't previously have a clear onchain use case (so it's emitted as an event).

secbajor: seems we'll have to align on a path forward to get this going. 

@RonTuretzky from my perspective the updates here to the YD are not multiplier logic, they are just adding relevant Cycle data (meaning a YD distribution Cycle) which the new voting streak multiplier uses to implement its own functions. 

if you see an alternate strategy, about  hooking into the multiplier when the user votes, can you clarify how that could be implemented? the only way i can see that working is also adding logic to the YD in the `castVote` function 🤔

RonTuretzky: I'm against adding anymore variables to the YD in general , it will require complex upgrade logic and add more variables I'm not sure are needed. 

```
My suggestion would be to hook into the multiplier when the user votes , update their last voted time and increment a counter. Next time they vote, we can use the CYCLE_LENGTH to determine whether they've broken their streak or not by referencing whether their last time voted is longer than CYCLE_LENGTH
``` 

^ this is my suggestion

bagelface: As per my previous comment, we need clarification from you on what this means:

> My suggestion would be to hook into the multiplier when the user votes"
> What exactly do you mean by this? How can we "hook" into the multiplier when the user votes without updating the YieldDistributor or adding an additional transaction? I guess we can add a contract that wraps the vote transaction and updates multipliers in the same transaction, but that just seems awkward. Feel free to DM me to discuss.

secbajor:
After some discussion and advice we are going a different direction. Closing in favor of https://github.com/BreadchainCoop/breadchain/pull/129

PR #129, opened 2025-04-24
Implement VotingStreakMultiplier

secbajor:
When a vote is cast (via the YieldDistributor), the VotingStreakMultiplier updates it's state variables tracking user's multiplier value and validity. 

When this contract is deployed, the front end can call `getMultiplyingFactor` for the user's current multiplier.


src/YieldDistributor.sol
```diff
@@ -240,6 +245,7 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         _currentVotingPower = multiplier == 0 ? _currentVotingPower : (_currentVotingPower * multiplier) / PRECISION;
         if (_currentVotingPower < minRequiredVotingPower) revert BelowMinRequiredVotingPower();
         _castVote(msg.sender, _points, _currentVotingPower);
+        votingStreakMultiplier.onVoteCast(msg.sender);
```
secbajor: @bagelface i could use some advice on how to update the tests, since adding this line causes failures in existing tests in VotingMultipliersTest in YieldDistributor.t.sol 

My understanding is that the YieldDistributor now needs to be initialized with the address of the VotingStreakMultiplier, but I imagine in the tests the idea is to mock that somehow instead of referencing a live contract. Can you point me in the right direction for how to do this?

src/YieldDistributor.sol
```diff
@@ -43,18 +44,20 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+    /// @notice The voting power allocated to each project by a specific voter in the current cycle
     mapping(address => uint256[]) voterDistributions;
     /// @notice How much of the yield is divided equally among projects
     uint256 public yieldFixedSplitDivisor;
     /// @notice The address of the `ButteredBread` token contract
     ERC20VotesUpgradeable public BUTTERED_BREAD;
     /// @notice The block number before the last yield distribution
     uint256 public previousCycleStartingBlock;
+    /// @notice The interface for the `VotingStreakMultiplier` contract
+    IVotingStreakMultiplier public votingStreakMultiplier;
```
bagelface: We don't want to add individual multipliers like this because (1) we can't because we can't call the initialize function again and (2) it's redundant because we're already maintaining a list in `VotingMultipliers`. Instead we should update or extend the functionality of `VotingMultipliers` and `IMultiplier` so that the `castVoteWithMultipliers` iterates over the multipliers and calls a new state-altering "update" function for any multiplier that implements that function. I think for the sake of simplicity, we can just generically update `IMultiplier` to include "update" in the interface so it's available for any multiplier and then the one's that don't need it can just no op

bagelface (changes requested):
needs slight redesign

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,113 @@
...
+
+        uint256 currentMultiplier = _getMultiplyingFactor(voter);
+        uint256 newMultiplier = (currentMultiplier == 0)
+            ? multiplierIncrement
+            : Math.min(currentMultiplier + multiplierIncrement, maxMultiplier * multiplierIncrement);
+
+        userToMultiplier[voter] = newMultiplier;
+        userToValidity[voter] = yieldDistributor.lastClaimedBlockNumber() + 2 * yieldDistributor.cycleLength();
+
+        emit MultiplierUpdated(voter, newMultiplier, userToValidity[voter]);
```
RonTuretzky: I don't see how we're keeping track of the streak here 

note that there is already a `accountToLastVoted` variable in the YD that can be used to see whether the user has broken their streak or not 

also i think we can change the `getMultiplyingFactor` to not be a view so this logic an be put in there, that way we don't need to introduce `onVoteCast` to the YD

RonTuretzky (changes requested):

src/interfaces/multipliers/IMultiplier.sol
```diff
@@ -2,6 +2,9 @@
 pragma solidity ^0.8.0;
 
 interface IMultiplier {
+    /// @notice Updates the voting multiplier for `_user`.
+    function updateMultiplyingFactor(address _user) external returns (uint256);
```
RonTuretzky: @bagelface should we use virtual here or something to allow people to not implement it? I forget what the right keyword is

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,114 @@
...
+            return;
+        }
+
+        uint256 currentMultiplier = getMultiplyingFactor(user);
+        uint256 newMultiplier = (currentMultiplier == 0)
+            ? multiplierIncrement
+            : Math.min(currentMultiplier + multiplierIncrement, maxMultiplier * multiplierIncrement);
+
+        userToMultiplier[user] = newMultiplier;
+        userToValidity[user] = yieldDistributor.lastClaimedBlockNumber() + 2 * yieldDistributor.cycleLength();
```
RonTuretzky: why is it 2 and not 1? 🤔

RonTuretzky (commented):
pattern is implemented as expected! looking very elegant!

src/interfaces/multipliers/IMultiplier.sol
bagelface: I'm 99% sure there isn't a way to do that in Solidity. If a function is defined in the interface, the inheriting contract has to implement it, even if it's just an no op implementation. We could potentially add a new interface like "IStatefulMultiplier.sol" that inherits from "IMultiplier.sol", but then we'd need the multipliers to implement [EIP-165](https://eips.ethereum.org/EIPS/eip-165) so the YieldDistributor is able to identify which interface a given contract is using.

RonTuretzky: what's the solution here then? inherit this function and implement it with an empty return every time? there's got to be a better solution

src/multipliers/VotingStreakMultiplier.sol
secbajor: i found this a bit confusing to think through, writing it out here so you can see my thinking:  

if i vote in cycle 10, my streak needs to be valid until the end of the next cycle (11) which is also the beginning of cycle 12. 

so if cycle 10 starts on block 1000 and i vote on block 1050, and a cycle is 100 blocks:
- if my streak is valid until 1100, i lose my streak at start of cycle 11 even though i voted in cycle 10
- if my streak is valid until 1200, i lose my streak at start of cycle 12 unless i voted in cycle 11

as a connected wallet on the front end app, i should be able to see an active multiplier during cycle 11 if i voted in cycle 10

RonTuretzky: Makes perfect sense, thanks!!

src/interfaces/multipliers/IMultiplier.sol
secbajor: is this what that would look like? is this stupid? 
https://github.com/BreadchainCoop/breadchain/pull/129/commits/3a761f16bf68037c58f05b5a9ed3aeb5f04571a3

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -66,9 +66,17 @@ contract VotingStreakMultiplier is Initializable, OwnableUpgradeable, IMultiplie
...
     /// @param user The address of the user
-    function updateMultiplyingFactor(address user) external override {
+    /// @return The current multiplying factor
+    function _getMultiplyingFactor(address user) external view returns (uint256) {
+        return getMultiplyingFactor(user);
+    }
+
+    /// @notice Updates the multiplying factor for a user
+    function updateMultiplyingFactor() external override {
+        address user = msg.sender;
```
RonTuretzky: I believe this will be the YD and not the user itself, right? could be wrong

src/interfaces/multipliers/IMultiplier.sol
RonTuretzky: Yes that's the only way I can think of doing it , I've seen this implementation pattern before so it seems like it may be unavoidable

src/multipliers/VotingStreakMultiplier.sol
secbajor: updated to pass in the user address instead of relying on this which i agree may be inaccurate

secbajor:
@bagelface @RonTuretzky with my latest commit here, the build is compiling without errors/warnings and the tests all pass. Can you advise on next steps?

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,131 @@
...
+        emit MultiplierUpdated(_user, newMultiplier, userToValidity[_user]);
+    }
+
+    /// @notice This is a no-op function that exists solely to implement the IMultiplier interface
+    /// @dev This function does nothing and always returns, as the actual multiplier update logic
+    ///      is handled by the parameterless updateMultiplyingFactor() function
+    /// @param _newMultiplyingFactor Unused parameter
+    function updateMultiplyingFactor(uint256 _newMultiplyingFactor) external pure override {
+        return;
+    }
```
RonTuretzky: I think we can remove this one from this contract and the upstream interfaces as well if we're not using it

RonTuretzky (changes requested):
LGTM so far , please add a `VotingStreakMultiplier.t.sol` file with unit tests

src/multipliers/VotingStreakMultiplier.sol
secbajor: i thought this was a function to be called externally in case we ever decide to update the multiplier value. like, if we start out thinking discord membership should give 1.05x but then later 1.1x or something, would this be the way to make that update?

secbajor: oh.. in which case i should update it to actually set the value. 

so, what do you think? does it make more sense to have this set the value in case we want to update it, or to remove as you originally suggested?

secbajor: i confused myself a bit but that's my thinking for why we need to keep it in the upstream interfaces, since those are implemented by other kinds of multipliers. and so my understanding is it has to be implemented here

RonTuretzky: I think basically you added one that was meant to be for this pr , then we figured out we need that one to accept address not an int, and now we have the previous one that's still there that we need to remove , no?

RonTuretzky: I still don't see where the int version of `updateMultiplyingFactor` is being used so bumping this for its removal

secbajor: @RonTuretzky the idea seems to be that it will be used by NFTMultiplier [here](https://github.com/BreadchainCoop/breadchain/pull/129/files#diff-6d674ebc7630d7430f73234b4b7ebd815a0bea5eb2f1c90fa28943e7e13a4454R66)

RonTuretzky: why would it not just be calculated and called internally? it being supplied implies that some outside contract is doing the logic to determine the multiplier update, but this is an anti pattern to the system design, as all multiplier logic is meant to be self contained within the multiplier contracts themselves

secbajor: okay cool -- sounds good to me :) 
i've gone ahead and removed it from the contract and interfaces

secbajor:
@RonTuretzky @bagelface 
I resolved ron's feedback. Functionality is in place for voting streak and i believe i added full test coverage. 

Can you advise now on next steps?

test/multipliers/VotingStreakMultiplier.t.sol
```diff
@@ -0,0 +1,338 @@
...
+import {YieldDistributorTestWrapper} from "src/test/YieldDistributorTestWrapper.sol";
+
+abstract contract Bread is ERC20VotesUpgradeable, Ownable2StepUpgradeable {
+    function claimYield(uint256 amount, address receiver) public virtual;
+    function yieldAccrued() external view virtual returns (uint256);
+    function setYieldClaimer(address _yieldClaimer) external virtual;
+    function mint(address receiver) external payable virtual;
+}
+
+contract VotingStreakMultiplierTest is Test {
```
RonTuretzky: There is already a Voting multiplier testing section [here](https://github.com/BreadchainCoop/breadchain/blob/952757e85598922624ddf0d9f4bb363c04e72f25/test/YieldDistributor.t.sol#L454) , would probably be better to refactor this to extend that area

not gonna block this PR on that though tbh, thoughts @bagelface ? 

Lets keep in mind this will all be refactored for breadkit anyways

test/multipliers/VotingStreakMultiplier.t.sol
```diff
@@ -0,0 +1,338 @@
...
+
+        // Roll to the next cycle
+        cycleIterator++;
+        setUpForCycle(yieldDistributor, cycleIterator);
+
+        castVote(testAccount);
+
+        // Verify multiplier does not exceed maxMultiplier * multiplierIncrement
+        assertEq(multiplier.getMultiplyingFactor(testAccount), MAX_MULTIPLIER * MULTIPLIER_INCREMENT);
+    }
```
RonTuretzky: These tests should be removed in favor of a fuzz test 

You can reference the two tests [here](https://github.com/BreadchainCoop/breadchain/blob/952757e85598922624ddf0d9f4bb363c04e72f25/test/YieldDistributor.t.sol#L321) for how you can turn a vanilla test into a fuzz test

RonTuretzky (changes requested):

test/multipliers/VotingStreakMultiplier.t.sol
secbajor: oh thats neat! thanks for the tip. 

replaced those tests with a fuzz test fe64fc65771c8e4083ce5b8700ffcb562ebc76da

test/multipliers/VotingStreakMultiplier.t.sol
secbajor: ok i moved them to the yd 53760537e198a5c33b7838e56d7efd2bb31c902a

RonTuretzky (approved):
All tests are passing, lgtm 

@bagelface any notes?

RonTuretzky:
@secbajor try to follow the upgrade safety validation steps in the README.md and lemme know if there are any issues

secbajor:
> @secbajor try to follow the upgrade safety validation steps in the README.md and lemme know if there are any issues

@RonTuretzky thanks for the direction; i did it but had to do [this](https://github.com/BreadchainCoop/breadchain/pull/129/commits/3c86cca0c99b3544aee1866d907e079daf23267b) to get there -- does it seem OK?

RonTuretzky:
> > @secbajor try to follow the upgrade safety validation steps in the README.md and lemme know if there are any issues
> 
> @RonTuretzky thanks for the direction; i did it but had to do [this](https://github.com/BreadchainCoop/breadchain/pull/129/commits/3c86cca0c99b3544aee1866d907e079daf23267b) to get there -- does it seem OK?

no, you should have used `forge flatten` instead of imports, which puts all the libraries into one file. This is to ensure that even if we upgrade libraries in the future, past versions won't create compilation conflicts. 

Please revert that commit and try using the `forge flatten` command instead , if you've done it correctly there will be no "import" lines in the file 

you may have to use `git submodule update` after the revert to reset your submodules

secbajor:
@RonTuretzky OK -- did that and now the latest commit is here f470adfa924df6f7a42b444c2f6fa27337f61fbf 

I included an update to the README that i think clarifies what the process is supposed to be, so please review to check my understanding

RonTuretzky:
> @RonTuretzky OK -- did that and now the latest commit is here [f470adf](https://github.com/BreadchainCoop/breadchain/commit/f470adfa924df6f7a42b444c2f6fa27337f61fbf)
> 
> I included an update to the README that i think clarifies what the process is supposed to be, so please review to check my understanding

lgtm, if upgrade safety validation ran with no issues we should be good 

@bagelface would still like you to take a look

RonTuretzky (approved):

src/multipliers/PermanentNFTMultiplier.sol
```diff
@@ -24,6 +24,7 @@ contract PermanentNFTMultiplier is INFTMultiplier {
     /// @notice Get the multiplying factor for a _user
     /// @param _user The address of the _user
     /// @return The multiplying factor if the _user has an NFT, 0 otherwise
+
```
bagelface: no line break

src/multipliers/NFTMultiplier.sol
```diff
@@ -60,10 +60,10 @@ contract NFTMultiplier is INFTMultiplier, Initializable, Ownable2StepUpgradeable
         return validUntilBlock;
     }
 
-    /// @notice Update the multiplying factor
-    /// @param _newMultiplyingFactor New multiplying factor (in basis points)
-    function updateMultiplyingFactor(uint256 _newMultiplyingFactor) external onlyOwner {
-        multiplyingFactor = _newMultiplyingFactor;
+    /// @notice Updates the multiplying factor for a specific user
+    /// @param _user The address of the user to update the multiplying factor for
+    function updateMultiplyingFactor(address _user) external view onlyOwner {
```
bagelface: Should remove the `onlyOwner` modifier here. Otherwise this will revert if anything but the owner calls it. The "YieldDistributor" contract likely won't be the owner.

test/YieldDistributor.t.sol
```diff
@@ -451,6 +454,190 @@ contract YieldDistributorTest is Test {
     }
 }
 
+contract VotingStreakMultiplierTest is YieldDistributorTest {
+    // uint256 constant START = 32_323_232_323;
+    uint256 constant MULTIPLIER_INCREMENT = 0.02e20;
+    uint256 constant MAX_MULTIPLIER = 3;
+
+    // YieldDistributorTestWrapper public yieldDistributor;
```
bagelface: Personally not a fan of leaving in commented out stuff like this. Either we need it or we don't. Commented out stuff creates ambiguity.

src/interfaces/multipliers/INFTMultiplier.sol
```diff
@@ -16,4 +16,8 @@ interface INFTMultiplier is IMultiplier {
     /// @param _user The address of the _user to check
     /// @return True if the _user owns at least one NFT, false otherwise
     function hasNFT(address _user) external view returns (bool);
+
+    /// @notice Updates the multiplying factor for a specific user
+    /// @param _user The address of the user to update the multiplying factor for
+    function updateMultiplyingFactor(address _user) external;
```
bagelface: This same abstract function is already declared in the `IMultiplier` interface that this interface inherits. I think this is redundant and be removed.

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,124 @@
...
+    uint256 public multiplierIncrement;
+
+    /// @notice The YieldDistributor contract
+    YieldDistributor public yieldDistributor;
+
+    /// @notice Mapping of user addresses to their current multiplier factor
+    mapping(address => uint256) public userToMultiplier;
+
+    /// @notice Mapping of user addresses to their multiplier validity period
+    mapping(address => uint256) public userToValidity;
```
bagelface: I would potentially rename this so it's more clear we're expecting a block number and how the validity is determined. Maybe "userToValidUntil" or "userToExpiresAt".

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,124 @@
...
+            return userToMultiplier[user];
+        }
+
+        return 0;
+    }
+
+    /// @notice Gets the current multiplying factor for a user
+    /// @param user The address of the user
+    /// @return The current multiplying factor
+    function _getMultiplyingFactor(address user) external view returns (uint256) {
```
bagelface: Functions with _ prefix are conventionally internal or private. I don't this function needs to exist at all actually since it's just a wrapper for an already public function.

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,124 @@
...
+
+        userToMultiplier[_user] = newMultiplier;
+        userToValidity[_user] = lastClaimedBlock + 2 * cycleLength;
+        emit MultiplierUpdated(_user, newMultiplier, userToValidity[_user]);
+    }
+
+    /// @notice Gets the validity period for a user's multiplier
+    /// @param user The address of the user
+    /// @return The block number until which the multiplier is valid
+    function validUntil(address user) external view override returns (uint256) {
```
bagelface: would bring this to the top with the other view functions

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,124 @@
...
+            return;
+        }
+
+        uint256 currentMultiplier = getMultiplyingFactor(_user);
+        uint256 newMultiplier = (currentMultiplier == 0)
+            ? multiplierIncrement
+            : Math.min(currentMultiplier + multiplierIncrement, maxMultiplier * multiplierIncrement);
+
+        userToMultiplier[_user] = newMultiplier;
+        userToValidity[_user] = lastClaimedBlock + 2 * cycleLength;
```
bagelface: add parentheses for readability
`userToValidity[_user] = lastClaimedBlock + (2 * cycleLength);`

test/YieldDistributor.t.sol
```diff
@@ -451,6 +454,190 @@ contract YieldDistributorTest is Test {
     }
 }
 
+contract VotingStreakMultiplierTest is YieldDistributorTest {
+    // uint256 constant START = 32_323_232_323;
+    uint256 constant MULTIPLIER_INCREMENT = 0.02e20;
```
bagelface: what's up with this number?

test/YieldDistributor.t.sol
```diff
@@ -451,6 +454,190 @@ contract YieldDistributorTest is Test {
...
+        address testAccount = address(0x1234);
+        address[] memory accounts = new address[](1);
+        accounts[0] = testAccount;
+        setUpAccountsForVoting(accounts);
+        return testAccount;
+    }
+
+    function setUpVotingStreakMultiplier() public returns (VotingStreakMultiplier) {
+        VotingStreakMultiplier multiplier = VotingStreakMultiplier(address(yieldDistributor.allowlistedMultipliers(0)));
+        multiplier.setMultiplierIncrement(MULTIPLIER_INCREMENT);
```
bagelface: this is already done when initializing the multiplier, no?

test/YieldDistributor.t.sol
```diff
@@ -451,6 +454,190 @@ contract YieldDistributorTest is Test {
...
+        multiplier.setMaxMultiplier(MAX_MULTIPLIER);
```
bagelface: same here

bagelface (approved):
have a few readability and style comments, but nothing that would prevent this from being merged in. Great stuff!

src/VotingMultipliers.sol
```diff
@@ -71,8 +80,8 @@ contract VotingMultipliers is Ownable2StepUpgradeable, IVotingMultipliers {
     /// @param _user The address of the user
     /// @param _multiplierIndexes Array of multiplier indexes to use
     /// @return The total multiplier value for the user
-    function getTotalMultipliers(address _user, uint256[] calldata _multiplierIndexes) public view returns (uint256) {
-        uint256 _totalMultiplier = 0;
+    function getTotalMultipliers(address _user, uint256[] calldata _multiplierIndexes) public returns (uint256) {
```
bagelface: get functions should have the view visibility. this won't be very useful on the frontend as is. i would remove `updateMultiplyingFactor` to allow this to be view. then just tolerate that it might be out of sync on the frontend. maybe add a "isSynced" flag based on timestamps and cycle length that can be checked and return alongside the multiplier value?

secbajor: @bagelface this function gets called from the yielddistributor, but the following function of the same name has a `@dev` notice that its intended to be called from front end, and is a `view` function. 

so my sense is this should remain as is, but please clarify if you see it differently.

bagelface: My issue with this change is mostly from a readability perspective. The function name is `getTotalMultipliers` but it is not a view function, which is inconsistent. The visibility should either be updated or the name of the function should change so that it's clear that state changes are being made when it's called.

src/interfaces/multipliers/IDynamicNFTMultiplier.sol
```diff
@@ -5,6 +5,7 @@ import {INFTMultiplier} from "src/interfaces/multipliers/INFTMultiplier.sol";
 /// @title Dynamic NFT Multiplier Interface
 /// @notice Interface for contracts that provide a dynamic multiplying factor for _users based on NFT ownership
 /// @dev Extends the INFTMultiplier interface with dynamic multiplier functionality
+
```
bagelface, line 8: no line break

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import {IMultiplier} from "src/interfaces/multipliers/IMultiplier.sol";
+import {YieldDistributor} from "src/YieldDistributor.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
```
bagelface: keep openzeppelin imports together

src/multipliers/VotingStreakMultiplier.sol
```diff
@@ -0,0 +1,130 @@
...
+
+/// @title VotingStreakMultiplier
+/// @notice A contract for managing voting streak multipliers
+/// @dev Implements IMultiplier and IVotingStreakMultiplier interfaces
+contract VotingStreakMultiplier is Initializable, OwnableUpgradeable, IMultiplier {
+    /// @notice The base multiplier value (100% in fixed-point representation)
+    uint256 constant BASE_MULTIPLIER = 1e18;
```
bagelface: usually good to be explicit with accessibility. this will default to `private`

src/interfaces/multipliers/IMultiplier.sol
```diff
@@ -2,7 +2,11 @@
 pragma solidity ^0.8.0;
 
 interface IMultiplier {
```
bagelface, line 6: I actually might be a good idea to just add `BASE_MULTIPLIER` constant here

test/YieldDistributor.t.sol
```diff
@@ -528,8 +721,11 @@ contract VotingMultipliersTest is YieldDistributorTest {
         vm.startPrank(voter);
         yieldDistributor.castVoteWithMultipliers(points, multiplierIndices);
 
-        // Expected voting power = initial * (1.5 + 2.0)
-        uint256 expectedVotingPower = (initialVotingPower * 3.5e18) / yieldDistributor.PRECISION();
+        // Expected voting power = initial * boost
+        // Start with base 100% (1e18) and add only the bonus amounts
+        uint256 totalMultiplier = 1e18; // Base 100%
+        totalMultiplier += (1.5e18 - 1e18) + (2e18 - 1e18); // Add bonuses from both multipliers
```
bagelface, line 727: just heads up, you can use the `ether` keyword to represent this like:

`totalMultiplier += (1.5 ether - 1 ether) + (2 ether - 1 ether);`

bagelface (approved):
a few comments, but lgtm

src/interfaces/multipliers/IMultiplier.sol
secbajor: i had this instinct as well, but found that solidity prevents variables from being set in interfaces. cursor suggested i create a library directory and define it, like MultiplierConstants, and import it from there. Do you think that's a good solution ?

secbajor: did this in a separate commit so its clearer f17990df7b02ad0fcb692585dfacc8ab283aa59f - wdyt?

test/YieldDistributor.t.sol
secbajor: im going to choose to leave it like 1e18 only because i think thats a little more self explanatory than `ether`, and i havent gotten the sense you have a strong preference

secbajor: good to know tho, appreciate the tip!!

src/interfaces/multipliers/IMultiplier.sol
bagelface: oh weird I didnt know that. I do think this makes things a little easier to work with so that the value doesn't need to be explicitly defined each time

PR #139, opened 2025-07-13
feat: add allowlistedMultipliers array getter in YieldDistributor and IYieldDistributor & tests

### LINKED ISSUE #126: Update VotingMultiplier with a function that returns all allowed multipliers
Background
The [current implementation](https://github.com/BreadchainCoop/breadchain/blob/952757e85598922624ddf0d9f4bb363c04e72f25/src/interfaces/IVotingMultipliers.sol#L24) of `allowListedMultipliers` requires passing in an index, and returning the allowed multiplier at the given index. To display all multipliers on the front-end app to a user, this would require N subsequent calls in order to display all available multipliers (boosts). Instead, I want to be able to query the contract once and retrieve a full list of multipliers.

Actions
1. Provide a new function in the VotingMultiplier interface that returns a list of the allowed multipliers. It will not have any arguments required.
2. Update tests as needed

lirona:
#126

### DIFF
--- a/src/VotingMultipliers.sol
+++ b/src/VotingMultipliers.sol
@@ -37,6 +37,13 @@ contract VotingMultipliers is Ownable2StepUpgradeable, IVotingMultipliers {
         return $.allowlistedMultipliers[index];
     }
 
+    /// @notice Returns the array of allowlisted multipliers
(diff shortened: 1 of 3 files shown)


RonTuretzky (approved):
lgtm

PR #174, opened 2025-11-13
Integrate GasKillerSDK into YieldDistributor

bagelface:
The `gas-killer-solidity-sdk` repo is private so running the CI will fail until we update the workflow with an access token. Figured we didn't want to block progress on this for that reason. You can verify tests pass locally with `forge test`.

Upgrade validation passes if you make some changes to trick the script into disregarding issues with initializers. We messed up the VotingMultipliers deploy, so I think it's a little confused. You can get it to pass if you add a call to `__VotingMultipliers_init` in the `initializeGasKiller` function.

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: GPL-3.0
-pragma solidity 0.8.25;
+pragma solidity ^0.8.25;
 
-import {ERC20VotesUpgradeable} from
-    "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import {
+    ERC20VotesUpgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
 import {EIP712Upgradeable} from "@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol";
 import {Ownable2StepUpgradeable} from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import {ReentrancyGuardUpgradeable} from
-    "@openzeppelin/contracts-upgradeable/utils/ReentrancyGuardUpgradeable.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 
 import {IButteredBread} from "src/interfaces/IButteredBread.sol";
 import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
@@ -20,7 +20,7 @@ import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
  * @custom:coauthor @daopunk
  * @custom:coauthor @bagelface
  */
-contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable {
+contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpgradeable, ReentrancyGuard {
     /// @notice Value used for calculating the precision of scaling factors
     uint256 public constant FIXED_POINT_PERCENT = 100;
     /// @notice `IERC20Votes` contract used for powering `ButteredBread` voting
@@ -52,7 +52,6 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpg
         __EIP712_init(_initData.name, "1");
         __ERC20Votes_init();
         __Ownable_init(msg.sender);
-        __ReentrancyGuard_init();
 
         for (uint256 i; i < _initData.liquidityPools.length; ++i) {
             scalingFactors[_initData.liquidityPools[i]] = _initData.scalingFactors[i];
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -21,8 +21,11 @@ import {VotingMultipliers} from "src/VotingMultipliers.sol";
  * @custom:coauthor prosalads.eth
  * @custom:coauthor kassandra.eth
  * @custom:coauthor theblockchainsocialist.eth
+ * @custom:coauthor cryptokibbutznik.eth
  * @custom:coauthor github.com/daopunk
  * @custom:coauthor github.com/secbajor
+ * @custom:coauthor github.com/hudsonhrh
+ * @custom:coauthor github.com/Tranquil-Flow
  */
 contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingMultipliers, GasKillerSDK {
     /// @notice The address of the $BREAD token contract
@@ -37,7 +40,7 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
     uint256 public minRequiredVotingPower;
     /// @notice The block number of the last yield distribution
     uint256 public lastClaimedBlockNumber;
-    /// @notice The total number of votes cast in the current cycle
+    /// @notice The total voting power accumulated in the current cycle
     uint256 public currentVotes;
     /// @notice Array of projects eligible for yield distribution
     address[] public projects;
@@ -57,6 +60,12 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
     ERC20VotesUpgradeable public BUTTERED_BREAD;
     /// @notice The block number before the last yield distribution
     uint256 public previousCycleStartingBlock;
+    /// @notice Array of voters who have cast votes in the current cycle
+    address[] public voters;
+    /// @notice The mapping of holders to their vote distributions
+    mapping(address => uint256[]) public holderToDistribution;
+    /// @notice The mapping of holders to their total vote distribution
+    mapping(address => uint256) public holderToDistributionTotal;
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() GasKillerSDK(address(0), address(0)) {
@@ -190,12 +199,11 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         uint256 _available_yield = BREAD.balanceOf(address(this)) + BREAD.yieldAccrued();
         if (
             /// No votes were cast
-            /// Already claimed this cycle
+            /// OR already claimed this cycle
+            /// OR yield is insufficient
             currentVotes == 0 || block.number < lastClaimedBlockNumber + cycleLength
                 || _available_yield / yieldFixedSplitDivisor < projects.length
         ) {
-            /// Yield is insufficient
-
             return (false, new bytes(0));
         } else {
             return (true, abi.encodePacked(this.distributeYield.selector));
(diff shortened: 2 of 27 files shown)


rubydusa:
Just curios why no CI/CD (not even the simple forge build) ran on this PR - guess the workflow is not configured to target merges into dev

I skimmed over the issues and the open PRs and I see #163 which seems to have gone stale, maybe for a good reason

src/ButteredBread.sol
```diff
@@ -20,7 +20,7 @@ import {IERC20Votes} from "src/interfaces/IERC20Votes.sol";
  * @custom:coauthor @daopunk
  * @custom:coauthor @bagelface
  */
-contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpgradeable, ReentrancyGuardUpgradeable {
+contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpgradeable, ReentrancyGuard {
```
rubydusa, line 23: why? This definitely does not look good because `ReentrancyGuard` has a storage slot.

src/YieldDistributor.sol
```diff
@@ -73,7 +83,6 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         uint256 _lastClaimedBlockNumber,
         address[] memory _projects
     ) public initializer {
-        VotingMultipliers.initialize();
```
rubydusa, line 76: why?
not that important since this PR is not intended for a new deployment but still - looks suspicious

test/upgrades/v1.0.3/ButteredBread.sol
rubydusa, line 1: I personally do not like linting on these type of "fixture" source files: IMO should be restored

src/YieldDistributor.sol
```diff
@@ -98,6 +107,26 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+    /**
+     * @notice Initializes the GasKiller SDK
+     * @param _avsAddress The address of the AVS service manager
+     * @param _blsSignatureChecker The address of the BLS signature checker
+     * @custom:oz-upgrades-validate-as-initializer
+     */
+    function initializeGasKiller(address _avsAddress, address _blsSignatureChecker) public reinitializer(2) {
+        _setAvsAddress(_avsAddress);
+        _setBlsSignatureChecker(_blsSignatureChecker);
+    }
```
rubydusa, line 126: I propose https://github.com/BreadchainCoop/gas-killer-solidity-sdk/pull/27

src/YieldDistributor.sol
```diff
@@ -178,51 +207,92 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+        uint256 _baseSplit,
+        uint256 _votedYield
+    ) internal {
+        // Execute transfers to projects
         for (uint256 i; i < projects.length; ++i) {
-            uint256 _votedSplit = ((projectDistributions[i] * _votedYield * PRECISION) / currentVotes) / PRECISION;
-            BREAD.transfer(projects[i], _votedSplit + _baseSplit);
+            uint256 _votedSplit = ((distributions[i] * _votedYield * PRECISION) / totalVotes) / PRECISION;
+            bool _success = BREAD.transfer(projects[i], _votedSplit + _baseSplit);
+            if (!_success) revert TransferFailed();
```
rubydusa, line 258: if one of the projects has a bad address does it mess up everything?

src/YieldDistributor.sol
```diff
@@ -178,51 +207,92 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
 
+    /**
+     * @notice Distribute $BREAD yield to projects using GasKiller voting system
+     */
+    function distributeYieldGK() public trackState {
+        (uint256 _balance, uint256 _baseSplit, uint256 _votedYield) = _claimAndPrepareYield();
+        (uint256[] memory _currentProjectDistributions, uint256 _totalVotes) = _commitVotedDistribution();
+
+        _executeAndFinalizeDistribution(_currentProjectDistributions, _totalVotes, _balance, _baseSplit, _votedYield);
+    }
```
rubydusa, line 278: This is the canonical function that GK references, right?
Shouldn't we disable calling it not through gas killer?

bagelface:
> Just curios why no CI/CD (not even the simple forge build) ran on this PR - guess the workflow is not configured to target merges into dev
> 
> I skimmed over the issues and the open PRs and I see #163 which seems to have gone stale, maybe for a good reason

This repo is set up to require manually triggering of CI/CD for some reason. We should fix that, I agree. For now, this PR won't pass any kind of CI because the first step of building will fail because it now has a dependency on a private repo https://github.com/BreadchainCoop/gas-killer-solidity-sdk

rubydusa (changes requested):

src/YieldDistributor.sol
bagelface: This causes a problem with integration into upgradeable contracts that use OZ v5+ because we end up with two different versions of the Initializable contract. We might be able to get it to work by playing around with the remappings in the gas killer repo (i.e. use OZ v5 gas killer contracts, allow OZ v4.9 for `eigenlayer-middleware` contracts) but it wasn't working out so this was used as a workaround for now.

test/upgrades/v1.0.3/ButteredBread.sol
bagelface: Personally, i think there's no reason to even maintain anything past the previous version of the contract. This needed to be changed because it's requiring a specific version of solc and we needed to bump to 0.8.27 for gas killer.

src/ButteredBread.sol
bagelface: in OZ v5.5.0, the ReentrancyGuard is moved to core library https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol but it still implements EIP-7201 so it's upgrade safe. idk why they did this because it's definitely alarming looking the way that it is, but I guess it's in an effort to avoid having to maintain a contract per repo that does essentially the same thing.

src/YieldDistributor.sol
bagelface: I'm just sort of cleaning up a mess with this change. The `VotingMultipliers` initializer was implemented incorrectly as it's own initializer and called directly, instead of:

```
/// @notice Initializes the contract
    /// @param _initialOwner The address of the initial owner
    function __VotingMultipliers_init(address _initialOwner) public onlyInitializing {
        __Ownable_init(_initialOwner);
    }
```

And having a new function with the `reinitializer` modifier. I moved it to it's own function for readability, but the function will get locked once we call `initializeGasKiller`.

I can change it back, I just thought this would look cleaner.

src/YieldDistributor.sol
bagelface: This rollout is just to get the functionality available for next cycle. We want to continue using it like normal for now so we kept both functions. Once we're fully migrated to gas killer with a few successful rounds we can update the contract to remove the old function.

src/YieldDistributor.sol
bagelface: Yea that is true. Is that a feature or a bug? I'm happy to change it back, but this seemed more robust

src/YieldDistributor.sol
rubydusa: yeah lol in my head read it as `__VotingMultipliers_init`

src/YieldDistributor.sol
rubydusa: I understand...
I still don't like that it's not clear how to initialize `GasKillerSDK` and it's documented in a github issue that you need to call `_setAvsAddress` and `_setBlsSignatureChecker` -
At the very least we need to document this in the source code

Another idea that popped into my head is something like `__GasKillerSDK_init_dirty` which is like `__GasKillerSDK_init` without the modifier with a comment above it explaining why we can't have `onlyInitializing`

src/YieldDistributor.sol
rubydusa: If theoretically this happens - will it be possible to remove the offending project and still have correct distributions (excluding the bad receiver contract)
@RonTuretzky

src/YieldDistributor.sol
rubydusa: It's not about removing the old function - it's about that the new function is not meant to be called directly

Are we sure that it can't be invoked in a way that messes things up?

src/YieldDistributor.sol
bagelface: In this case we'd need to update the contract to include a "rescue" function that would let us call `_updateBreadchainProjects` via a public function directly (rather than by calling `_executeAndFinalizeDistribution`). It's not an ideal scenario, but definitely something we can work around if it happens.

src/YieldDistributor.sol
bagelface: It can be called directly if you a user wants, just like the original `distributeYield` function. There are checks within the function so that it won't distribute if it's not a valid time to distribute. The GK function is just expensive in a scenario where there is a large amount of voters, so we have the gas killer `verifyAndUpdate` function for that.

src/YieldDistributor.sol
bagelface: Ya I agree, it's not ideal. I can definitely open an issue on the `gas-killer-solidity-sdk` repo to add documentation on proper usage. I'm not against your suggestion, but want to take another shot at getting it to work properly first. We're definitely cutting some corners here, but the idea is to get the YD updated for this cycle so we can use gas killer for the next distribution.

test/upgrades/v1.0.3/ButteredBread.sol
RonTuretzky: please make your comments on https://github.com/BreadchainCoop/etherform/pull/3 , this is out of scope of this pr

src/YieldDistributor.sol
RonTuretzky: @bagelface @rubydusa this is off scope and its an acceptable concession that the multisig voters have to validate the address is valid before they sign the tx. off scope for this pr as we're not trying to improve the trust model of the yield distribution system. opening an issue in the breadkit repo for this to make sure it's implemented there.

RonTuretzky: https://github.com/BreadchainCoop/crowdstake.fun/issues/80

src/YieldDistributor.sol
RonTuretzky: @rubydusa it is in fact by design meant to be able to be called directly - this is part of the gas killer value proposition - backwards compatiblity

src/YieldDistributor.sol
```diff
@@ -20,12 +19,15 @@ import {VotingMultipliers} from "src/VotingMultipliers.sol";
  * @custom:coauthor prosalads.eth
  * @custom:coauthor kassandra.eth
  * @custom:coauthor theblockchainsocialist.eth
+ * @custom:coauthor cryptokibbutznik.eth
```
RonTuretzky:
```suggestion
 * @custom:coauthor postcapitalistcrypto.eth
```

src/YieldDistributor.sol
```diff
@@ -403,7 +523,16 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+    function setButteredBread(address _butteredBread) public onlyOwner trackState {
+        BUTTERED_BREAD = IERC20Votes(_butteredBread);
+    }
+
+    /**
+     * @notice Allows the owner to set the AVS address
+     * @param newAvsAddress The new AVS address
+     * @dev Also updates the namespace for the contract
+     */
+    function setAvsAddress(address newAvsAddress) external onlyOwner trackState {
```
RonTuretzky, line 535: we need a method to set the bls signature checker address not the avs address as we dont use that right now

RonTuretzky (changes requested):

src/YieldDistributor.sol
```diff
@@ -73,7 +81,6 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
         uint256 _lastClaimedBlockNumber,
         address[] memory _projects
     ) public initializer {
```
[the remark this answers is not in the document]
RonTuretzky: @bagelface we will have to re-init the voting mutlipliers?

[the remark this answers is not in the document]
bagelface: No, it's already been initialized. This is sort of just a readability clean up so we can better retrace our upgrade history and so I could make the change to the "VotingMultipliers" contract to remove the public `initializer` function an replace it with the internal `__VotingMultipliers_init` which is more appropriate for a contract that's designed to be inherited by another contract ("YieldDistributor").

The `initializeVotingMultipliers` will be locked when we call `initializeGasKiller` because the reinitializer index for the later function is > than the former.

src/YieldDistributor.sol
```diff
@@ -178,51 +205,92 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+        uint256 _baseSplit,
+        uint256 _votedYield
+    ) internal {
+        // Execute transfers to projects
         for (uint256 i; i < projects.length; ++i) {
-            uint256 _votedSplit = ((projectDistributions[i] * _votedYield * PRECISION) / currentVotes) / PRECISION;
-            BREAD.transfer(projects[i], _votedSplit + _baseSplit);
+            uint256 _votedSplit = ((distributions[i] * _votedYield * PRECISION) / totalVotes) / PRECISION;
+            bool _success = BREAD.transfer(projects[i], _votedSplit + _baseSplit);
+            if (!_success) revert TransferFailed();
```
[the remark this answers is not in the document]
bagelface: IERC20Votes extends IERC20 which implements `transfer() returns bool` as expected by this function

RonTuretzky (approved):

PR #176, opened 2025-11-30
Remove minimum voting power requirement

RonTuretzky:
## Summary
- Remove `minRequiredVotingPower` state variable and related checks in `castVote` and `castVoteWithMultipliers`
- Remove `BelowMinRequiredVotingPower` error from the interface
- Remove `setMinRequiredVotingPower` admin function
- Add `_initialOwner` parameter to `initialize` function and call `__VotingMultipliers_init` to fix initialization (tests were failing because ownership wasn't set up)
- Update deploy scripts and test configs to remove the deprecated parameter

This allows all BREAD holders to vote regardless of their balance, fixing the issue where `getVotingPower` returns 0 even when users have some voting power but below the minimum threshold.

## Test plan
- [x] All existing tests pass (80 tests)
- [x] Removed test for `BelowMinRequiredVotingPower` revert
- [x] Verified voting works without minimum threshold

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,13 +2,7 @@ name: test
 
 on:
   push:
-    branches:
-      - main
-      - dev
   pull_request:
-    branches:
-      - main
-      - dev
   workflow_dispatch:
 
 jobs:
--- a/script/deploy/DeployYieldDistributor.s.sol
+++ b/script/deploy/DeployYieldDistributor.s.sol
@@ -13,7 +13,6 @@ contract DeployYieldDistributor is Script {
     string config_data = vm.readFile(deployConfigPath);
     address _bread = stdJson.readAddress(config_data, "._bread");
     address _butteredBread = stdJson.readAddress(config_data, "._butteredBread");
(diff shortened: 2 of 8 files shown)


bagelface:
How would you feel about instead, just removing the `MustBeGreaterThanZero` checks for minimum voting power and just setting the value to 0?

RonTuretzky:
> How would you feel about instead, just removing the `MustBeGreaterThanZero` checks for minimum voting power and just setting the value to 0?

its a rough and dirty workaround, we would have to backlog a fix, but sure

bagelface:
> > How would you feel about instead, just removing the `MustBeGreaterThanZero` checks for minimum voting power and just setting the value to 0?
> 
> its a rough and dirty workaround, we would have to backlog a fix, but sure

Why is this a "rough and dirty workaround"? It's just making use of the existing functionality and removing the conditional we set preventing the state we want.

EDIT: Spoke directly with Ron. We need to remove the min voting power entirely because it prevents casting a vote for newly minted BREAD. The legacy path in the "_castVote" function needs to write voting power to storage. When using legacy distro, it will calcualted based on previous cycle window. When using current distro, we calculate the voting power after cycle is bumped, so it's OK for the voting power to be zero. It will still be counted.

PR #177, opened 2025-11-30
fix: include current cycle in getCurrentVotingPower

RonTuretzky:
## Summary
- Fixed `getCurrentVotingPower` to include voting power from the current cycle, not just the previous cycle
- Changed the end block parameter from `lastClaimedBlockNumber` to `block.number`

This resolves the issue where users who hold BREAD in the current cycle would see 0 voting power returned.

## Test plan
- [ ] Verify `getCurrentVotingPower` returns non-zero for users with BREAD in current cycle
- [ ] Verify existing voting functionality still works correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -139,8 +139,8 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
      * @return uint256 The voting power of the user
      */
     function getCurrentVotingPower(address _account) public view returns (uint256) {
-        return this.getVotingPowerForPeriod(BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account)
-            + this.getVotingPowerForPeriod(BUTTERED_BREAD, previousCycleStartingBlock, lastClaimedBlockNumber, _account);
+        return this.getVotingPowerForPeriod(BREAD, previousCycleStartingBlock, block.number, _account)
+            + this.getVotingPowerForPeriod(BUTTERED_BREAD, previousCycleStartingBlock, block.number, _account);
     }
 
     /// @notice Get the current accumulated voting power for a user


RonTuretzky:
reccomending merging this in and setting the minimum voting power to 0 to allow for instant voting @bagelface 

we wil need to do another upgrade

bagelface:
I understand the thought behind this change, but I'm not sure if I like it to be honest. With this change each voter's voting power is calculated from the previous cycle window plus a continuously growing window for the current cycle. So every user is incentivized to vote as close to the close of the window as possible for the optimal amount of voting power. Personally, I think keeping it simpler is better. Was this change something that was discussed with the rest of the team?

RonTuretzky:
> I understand the thought behind this change, but I'm not sure if I like it to be honest. With this change each voter's voting power is calculated from the previous cycle window plus a continuously growing window for the current cycle. So every user is incentivized to vote as close to the close of the window as possible for the optimal amount of voting power. Personally, I think keeping it simpler is better. Was this change something that was discussed with the rest of the team?

I didn't mean to keep both last month and this months. 
This should be used in the greedy GK distro func where it uses this function at the time of distribution, so there is no misincentive of needing to vote at a specific time. 

This change is essentially to allow real time voting using gas killer

bagelface:
> > I understand the thought behind this change, but I'm not sure if I like it to be honest. With this change each voter's voting power is calculated from the previous cycle window plus a continuously growing window for the current cycle. So every user is incentivized to vote as close to the close of the window as possible for the optimal amount of voting power. Personally, I think keeping it simpler is better. Was this change something that was discussed with the rest of the team?
> 
> I didn't mean to keep both last month and this months. This should be used in the greedy GK distro func where it uses this function at the time of distribution, so there is no misincentive of needing to vote at a specific time.
> 
> This change is essentially to allow real time voting using gas killer

From what I can tell, this change doesn't need to happen to allow for real time voting. When we call GK distribution, the cycle is bumped creating a bounded window for that cycle (`_claimAndPrepareYield()`). After that window is created, we run `_commitVotedDistribution()` which calculates the voting power in that window.

It would only have an affect in the case of a legacy distribution, where the voting power is calculated when the vote is cast.

RonTuretzky:
> > > I understand the thought behind this change, but I'm not sure if I like it to be honest. With this change each voter's voting power is calculated from the previous cycle window plus a continuously growing window for the current cycle. So every user is incentivized to vote as close to the close of the window as possible for the optimal amount of voting power. Personally, I think keeping it simpler is better. Was this change something that was discussed with the rest of the team?
> > 
> > 
> > I didn't mean to keep both last month and this months. This should be used in the greedy GK distro func where it uses this function at the time of distribution, so there is no misincentive of needing to vote at a specific time.
> > This change is essentially to allow real time voting using gas killer
> 
> From what I can tell, this change doesn't need to happen to allow for real time voting. When we call GK distribution, the cycle is bumped creating a bounded window for that cycle (`_claimAndPrepareYield()`). After that window is created, we run `_commitVotedDistribution()` which calculates the voting power in that window.
> 
> It would only have an affect in the case of a legacy distribution, where the voting power is calculated when the vote is cast.

it is still not possible to , in our current version, mint bread and immediately vote. However you want to frame the change needed to allow that, my motivation is to allow for instant mint + vote. IMO the most coherent way to do this is to move off of historical voting power and calculate voting power at the time of distro. currently if you mint then try to vote the getCurrentVotingPower function will revert your vote.

bagelface:
>

Ok I see because of the minimum voting power restriction, it's blocking votes for newly minted bread. The simplest fix for this would be to just do https://github.com/BreadchainCoop/breadchain/pull/176. Making the change in this PR to set the limit using `block.number` has an impact on the legacy voting mechanism that I don't think we want.

bagelface:
"issue where users who hold BREAD in the current cycle would see 0 voting power returned" is resolved by https://github.com/BreadchainCoop/breadchain/pull/176

PR #180, opened 2026-02-05
Add voting cycles for more gas efficient distribution

bagelface:
The ultimate goal of these changes is to optimize the distribution functions so that less gas is spent on "unoptimizable" operations (in context of gas savings from Gas Killer). The key optimizations here are:

- remove `delete holderToDistribution[_voter];` in "_commitVotedDistribution()" used by "distributeYieldGK()"
- remove `delete holderToDistributionTotal[_voter];` in "_commitVotedDistribution()" used by "distributeYieldGK()"
- remove `delete voters;` in "_executeAndFinalizeDistribution()" used by both distribution functions

These are eliminated by a redesign where we track a voting cycle index. When voters vote, we set their voting cycle index to the current one and update a couple of mappings instead of pushing them onto an array. This way we can just increment the cycle index and reset the voter count.

NOTE: I'm not sure exactly how the frontend makes use of the "voters" variable. It may be the case that we need to update some frontend code.

### DIFF
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -140,10 +140,12 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
 
     /**
      * @notice Initializes the voting cycle
-     * @dev This is required for proxy deployments since storage defaults don't apply
+     * @param _votingCycle The voting cycle number to initialize
      * @custom:oz-upgrades-validate-as-initializer
      */
     function initializeVotingCycle(uint256 _votingCycle) public reinitializer(3) onlyOwner {
+        if (_votingCycle == 0) revert MustBeGreaterThanZero();
+
         if (votingCycle == 0) {
             // Only initialize if voting cycle is not already initialized
             _initializeVotingCycle(_votingCycle);
--- a/src/YieldDistributor.sol
+++ b/src/YieldDistributor.sol
@@ -61,9 +61,9 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
     /// @dev DEPRECATED: Kept for storage layout compatibility. Use `voterAtIndex` and `votersCount` instead.
     address[] public voters;
     /// @notice The mapping of holders to their vote distributions
-    mapping(address => uint256[]) public holderToDistribution;
+    mapping(address => uint256[]) internal _holderToDistribution;
     /// @notice The mapping of holders to their total vote distribution
-    mapping(address => uint256) public holderToDistributionTotal;
+    mapping(address => uint256) internal _holderToDistributionTotal;
     /// @notice The current voting cycle number (incremented each distribution)
     uint256 public votingCycle;
     /// @notice The number of voters in the current cycle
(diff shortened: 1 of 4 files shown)


src/YieldDistributor.sol
```diff
@@ -58,11 +58,20 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
     /// @notice The block number before the last yield distribution
     uint256 public previousCycleStartingBlock;
     /// @notice Array of voters who have cast votes in the current cycle
+    /// @dev DEPRECATED: Kept for storage layout compatibility. Use `voterAtIndex` and `votersCount` instead.
     address[] public voters;
     /// @notice The mapping of holders to their vote distributions
     mapping(address => uint256[]) public holderToDistribution;
```
bagelface: wondering if this and `holderToDistributionTotal` should be made internal and wrapped in getter functions, so that we can revert if address is queried that has not voted in the active cycle. That way we won't return stale data.

Example code:

```solidity
mapping(address => uint256[]) internal _holderToDistribution;
mapping(address => uint256) internal _holderToDistributionTotal;

function holderToDistribution(address _voter) public returns (uint256[]) {
  if (voterVotedCycle[_voter] == votingCycle) {
    return _holderToDistribution[_voter];
  }

  revert VoterHasNotVoted();
}

function holderToDistributionTotal(address _voter) public returns (uint256) {
  if (voterVotedCycle[_voter] == votingCycle) {
    return _holderToDistributionTotal[_voter];
  }

  revert VoterHasNotVoted();
}
```

RonTuretzky: what is the case where this happens?

bagelface: It's really a matter of how you look at it. If a voter's previous cycle vote distribution should be considered valid for the next one automatically (rolled over). It technically doesn't. They're required to explicitly revote for their vote to be counted. So returning these values for previous cycles is technically stale data. It's honestly pretty inconsequential. Could even be framed as a feature as it gives the frontend a way to "look back" to autofill their vote from the previous cycle.

hudsonhrh: I think we should make this chnage. dont think we need to preserve this for any frontend use. indexer means we can look back at previous voting cycles and who voted in them because we have events for voting logic

hudsonhrh (changes requested):
lgtm besides that comment

src/interfaces/IYieldDistributor.sol
```diff
@@ -23,12 +23,14 @@ interface IYieldDistributor {
     error ProjectNotFound();
     /// @notice The error emitted when attempting to calculate voting power for a period with a start block greater than the end block
     error StartMustBeBeforeEnd();
+    /// @notice The error emitted when a transfer fails
+    error TransferFailed();
```
bagelface, line 27: fixed alphabetical order

src/YieldDistributor.sol
```diff
@@ -58,11 +58,20 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
     uint256 public previousCycleStartingBlock;
     /// @notice Array of voters who have cast votes in the current cycle
+    /// @dev DEPRECATED: Kept for storage layout compatibility. Use `voterAtIndex` and `votersCount` instead.
     address[] public voters;
     /// @notice The mapping of holders to their vote distributions
-    mapping(address => uint256[]) public holderToDistribution;
+    mapping(address => uint256[]) internal _holderToDistribution;
     /// @notice The mapping of holders to their total vote distribution
-    mapping(address => uint256) public holderToDistributionTotal;
+    mapping(address => uint256) internal _holderToDistributionTotal;
```
rubydusa, line 66: why change to internal?

src/YieldDistributor.sol
```diff
@@ -441,6 +500,18 @@ contract YieldDistributor is IYieldDistributor, Ownable2StepUpgradeable, VotingM
...
+    /**
+     * @notice Internal function to initialize the voting cycle
+     * @dev Initialize voting cycle to 1 so that uninitialized voterVotedCycle mappings (default 0)
+     * @dev are correctly identified as "not voted in current cycle"
+     */
+    function _initializeVotingCycle(uint256 _votingCycle) internal {
+        votingCycle = _votingCycle;
+
+        // Reset deprecated `voters` array
+        delete voters;
```
rubydusa: why? this would create a one-time edge case where voters were reset not on distribution.

src/YieldDistributor.sol
bagelface: see discussion here https://github.com/BreadchainCoop/breadchain/pull/180#discussion_r2770516256

src/YieldDistributor.sol
bagelface: It's mostly just to keep "voters" from getting accessed and not having context that it's unused and unmaintained data once we upgrade to the voting cycle design. Now that I'm thinking about it, it might just be cleaner to change its visibility to private.

rubydusa (approved):

hudsonhrh (approved):
lgtm

PR #182, opened 2026-02-12
Refactor upgrade scripts and docs

bagelface:
- Remove old flattened contracts. There's not really any point in keeping anything but the latest. We can always recreate them if we need them. It just adds noise to the linter output.
- Create dedicated README for upgrade process so it's not all kept in the top level README


rubydusa (commented):
I like the latest tag addition but why remove tracking of versions in the upgrades directory? I think it serves as a nice reference

bagelface:
> I like the latest tag addition but why remove tracking of versions in the upgrades directory? I think it serves as a nice reference

I don't know if they're very helpful to get any kind of "at a glance" diff because they're flattened. Instead, you'd just come and look at the tag in the repo and do a compare. We also don't have any practical need to validate updates against them. And lastly, they create linter noise when building the contracts because a lot of them use older versions of OZ contracts. I've actually had to make a few manual changes in the past for them to build properly, so some of them aren't even the original output from the flatten utility.

rubydusa:
> > I like the latest tag addition but why remove tracking of versions in the upgrades directory? I think it serves as a nice reference
> 
> I don't know if they're very helpful to get any kind of "at a glance" diff because they're flattened. Instead, you'd just come and look at the tag in the repo and do a compare. We also don't have any practical need to validate updates against them. And lastly, they create linter noise when building the contracts because a lot of them use older versions of OZ contracts. I've actually had to make a few manual changes in the past for them to build properly, so some of them aren't even the original output from the flatten utility.

Fair enough. Approving

rubydusa (approved):

RonTuretzky:
sorry for the work that was done here but I would push this out until [this](https://github.com/BreadchainCoop/etherform) repo is in a workable state and then transition to it

bagelface:
> sorry for the work that was done here but I would push this out until [this](https://github.com/BreadchainCoop/etherform) repo is in a workable state and then transition to it

Can you elaborate on why? The repo you linked is a bunch of docs and some github workflows. If you look at these changes, you can see it's really nothing more than a cleanup. I update the READMEs and remove old flattened contracts.

RonTuretzky:
> > sorry for the work that was done here but I would push this out until [this](https://github.com/BreadchainCoop/etherform) repo is in a workable state and then transition to it
> 
> Can you elaborate on why? The repo you linked is a bunch of docs and some github workflows. If you look at these changes, you can see it's really nothing more than a cleanup. I update the READMEs and remove old flattened contracts.

👍

PR #191, opened 2026-03-13
Set default 'lastClaimedBlockNumber' to current block

bagelface:
In the deploy scripts, allow for setting the value 0 for `lastClaimedBlockNumber` so it will default to the current block (or close). This makes sure new deployments are starting their cycle at/near the block they are deployed at. It's imprecise, because the contract deployment might actually clear a block or two later, but we avoiding adding it as contract logic so we don't bloat the initializer.

This PR also updates the config json to include the real ButteredBread contract address instead of the placeholder.


hudsonhrh:
wouldnt it be a really tiny change to the initializer that ensures consistency? idk if bloating the initializer is a concern

hudsonhrh (commented):
left comment

bagelface:
> wouldnt it be a really tiny change to the initializer that ensures consistency? idk if bloating the initializer is a concern

Ya you right

PR #192, opened 2026-03-13
Improve mock deploy script

bagelface:
Minor changes to improve mock deploy script:
- Better handling of private key argument (allow for CLI flag or envar)
- Allow for specifying "owner" address as an argument
- Use default lastClaimedBlockNumber (deployed block)
- Reduce cycle length to 5 mins so it's easier to work with

### DIFF
--- a/script/deploy/test/DeployMockYieldDistributor.s.sol
+++ b/script/deploy/test/DeployMockYieldDistributor.s.sol
@@ -9,6 +9,10 @@ import {MockBread} from "src/test/MockBread.sol";
 import {ButteredBread, IButteredBread} from "src/ButteredBread.sol";
 import {YieldDistributor} from "src/YieldDistributor.sol";
 
+interface IVmWallets {
+    function getWallets() external returns (address[] memory);
+}
+
 contract DeployMockYieldDistributor is Script {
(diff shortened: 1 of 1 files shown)


script/deploy/test/DeployMockYieldDistributor.s.sol
```diff
@@ -9,6 +9,10 @@ import {MockBread} from "src/test/MockBread.sol";
 import {ButteredBread, IButteredBread} from "src/ButteredBread.sol";
 import {YieldDistributor} from "src/YieldDistributor.sol";
 
+interface IVmWallets {
```
hudsonhrh, line 12: just curious what this. is get wallets a built in function?

hudsonhrh (approved):
lgtm but had a question left comment about just curious

script/deploy/test/DeployMockYieldDistributor.s.sol
bagelface: basically interface for this https://foundry-rs.github.io/foundry/foundry_cheatcodes/Vm/struct.getWalletsCall.html but it's not available in every foundry version so that's why it's wrapped in try...catch

PR #193, opened 2026-03-18
Add gasKillerManager role for GasKiller configuration

RonTuretzky:
## Summary

Add a dedicated `gasKillerManager` role that can call `setAvsAddress`, `setBlsSignatureChecker`, and `setBlockStaleMeasure`. The contract owner retains the ability to call these functions directly and can delegate GasKiller configuration by setting the manager address.

## Changes

- Add `gasKillerManager` storage variable
- Add `onlyGasKillerManager` modifier allowing either the manager or owner
- Add `setGasKillerManager()` function for owner to assign the manager role
- Update three GasKiller setter functions to use the new access control

## Test plan

- Verify owner can still call the three setter functions
- Verify gasKillerManager can call the three setter functions
- Verify non-manager non-owner addresses are rejected


bagelface:
not needed

PR #208, opened 2026-08-10
fix: burn the difference of scaled balances so full LP withdrawals clear

bagelface:
Fixes the withdrawal failures reported in BreadchainCoop/crowdstaking-v2#408, where users could not unlock their LP tokens.

## Problem

`_deposit` floored its mint once per deposit, while `_withdraw` floored over the aggregate balance:

```solidity
// _deposit  →  floor() applied once per deposit
_mint(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);

// _withdraw →  floor() applied to the total
_burn(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
```

A sum of floors can fall short of the floor of the sum. So for any account that deposited **more than once**, withdrawing the full balance tried to burn more `ButteredBread` than was ever minted to them, and `_burn` reverted with `ERC20InsufficientBalance`.

The frontend unlocks the full `accountToLPBalance`, which is exactly the amount that reverts — so affected users could never succeed, regardless of wallet. Wallets that simulate before prompting (Safe, Rabby) refuse to open or hang on "in process", which is how this surfaced in the issue.

**Why it went unnoticed:** the live BREAD/WXDAI pool has a scaling factor of `32`, which is not a multiple of `FIXED_POINT_PERCENT`, so every deposit truncates. The existing suites only exercise `XDAI_FACTOR = 700`, an exact multiple of 100 where every division is exact and the shortfall cannot appear.

### Live example — `0x37cc13650d0D2A73E181cbF48A1847AE5f0531b5`

| | LP amount | BB minted | truncated |
|---|---|---|---|
| deposit @ block 38022377 | 2146701100755580030478 | 686944352241785609752 | 96 |
| deposit @ block 38625159 | 485633753042731023542 | 155402800973673927533 | 44 |
| **total** | **2632334853798311054020** | **842347153215459537285** | 140 → **1 unit lost** |

`floor(2632334853798311054020 × 32 / 100) = 842347153215459537286` — one wei more BB than exists.

**3 of 17 active depositors are currently affected** (~3031.95 BUTTER). The other 14 have no shortfall yet but would hit this on their next deposit. No funds were ever lost — staked balances still sum exactly to the LP the contract holds.

## Fix

Both sides now move by the **difference of scaled balances** rather than scaling the delta, so an account's holding stays equal to its scaled staked balance after every operation and a full exit always burns exactly what was minted:

```solidity
// _deposit
_mint(_account, _scaledBalance(newBalance, _lp) - _scaledBalance(previousBalance, _lp));

// _withdraw
_burnUpToBalance(_account, _scaledBalance(previousBalance, _lp) - _scaledBalance(newBalance, _lp));
```

`_syncVotingWeight` had the same floor-of-difference flaw and now scales each side before subtracting.

Burns are capped at the account's balance via `_burnUpToBalance`. This is what lets the **already-affected positions** exit: delta arithmetic alone keeps their pre-existing shortfall intact, so without the cap the three accounts above would still revert. For any deposit made under the new accounting the requested amount is already covered, so the cap never applies.

**No storage changes** — `bread`, `allowlistedLPs`, `scalingFactors`, and `_accountToLPData` keep their slots, so the upgrade is layout-safe and needs no migration.

## Testing

`104/104` passing. New coverage:

**`ButteredBreadTest_WithdrawRounding`** — replicates the incident against a fresh proxy:
- `testWithdrawFullBalanceAfterTwoDeposits` uses the exact on-chain deposit amounts at factor 32. Before the fix it reproduced the mainnet revert byte-for-byte: `ERC20InsufficientBalance(..., 842347153215459537285, 842347153215459537286)`.
- `testWithdrawFullBalanceFractionalFactor` shows the bug is **not** specific to the sub-100 factor — it also occurs at `132`, which satisfies the documented `>= FIXED_POINT_PERCENT` invariant.
- `testMintedSupplyMatchesAggregateEntitlement`, `testDustWithdrawalDoesNotRetainVotingWeight`, and two fuzz tests over arbitrary deposit/withdrawal splits and factors 1–10000.

**`ButteredBreadUpgradeTest`** — forks Gnosis at block 47651977 and upgrades the **live proxy**:
- `testFullWithdrawRevertsBeforeUpgrade` confirms all three accounts genuinely revert today.
- `testFullWithdrawSucceedsAfterUpgrade` confirms each recovers its **entire** staked balance in a **single call**.
- `testUpgradePreservesExistingState` asserts total supply, LP holdings, scaling factor, and every position are untouched by the upgrade.

All five previously-failing regression tests fail on `dev` and pass here.

### One pre-existing assertion changed

`ButteredBreadTest_Fuzz.testWithdrawAndBurn` asserted `balanceOf == preWithdrawBalance - floor(withdrawal × factor / 100)`, which encodes the old arithmetic and is off by one under the fix. It now asserts the invariant directly:

```solidity
assertEq(bb.balanceOf(ALICE), remainingBalance * _s.initialFactor / fixedPointPercent);
```

This is strictly stronger — it pins the holding to the scaled remainder rather than to a delta.

## Recovery without waiting for this upgrade

Affected users can exit today by withdrawing just under the max, then the dust. Verified on a fork; each recovers 100%:

```
0x37cc…31b5:  withdraw(lp, 2632334853798311054018)  then  withdraw(lp, 2)
0x18a7…8879:  withdraw(lp, 295991869687105715003)   then  withdraw(lp, 3)  then  withdraw(lp, 1)
0xcf26…1bf1:  withdraw(lp, 103620399017327657862)   then  withdraw(lp, 2)
```

Step 1 is `min(staked, (bb_balance × 100 + 99) / 32)`. Follow-up chunks must be ≤ 3 wei each, since `floor(4 × 32 / 100) = 1` with no BB left to burn.

## Left out of scope — worth separate decisions

1. **Is `32` the intended factor?** It means depositors receive **0.32×**, contradicting the documented `100 = 1X, 150 = 1.5X` semantics — possibly a missing zero. `initialize` has no `>= FIXED_POINT_PERCENT` guard while `modifyScalingFactor` does, so production sits at a value it can no longer be set back to. This PR makes the arithmetic correct for *any* factor and deliberately does not change the economics or add the guard, since that's a policy call.

2. **`modifyAllowList(lp, false)` also gates `withdraw`.** Delisting an LP would freeze every existing position in it. Dropping `onlyAllowed` from the withdraw path would keep exits open unconditionally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ButteredBread.sol
+++ b/src/ButteredBread.sol
@@ -150,23 +150,31 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpg
 
         /// @dev ensure proper accounting in case of admin error in `modifyScalingFactor` where not all holders are updated
         _syncVotingWeight(_account, _lp);
-        _accountToLPData[_account][_lp].balance += _amount;
 
-        _mint(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
+        uint256 previousBalance = _accountToLPData[_account][_lp].balance;
+        uint256 newBalance = previousBalance + _amount;
+        _accountToLPData[_account][_lp].balance = newBalance;
+
+        _mint(_account, _scaledBalance(newBalance, _lp) - _scaledBalance(previousBalance, _lp));
 
         emit ButterAdded(_account, _lp, _amount);
     }
 
     /// @notice Withdraw LP tokens and burn ButteredBread with corresponding LP scaling factor
     function _withdraw(address _account, address _lp, uint256 _amount) internal {
-        if (_amount > _accountToLPData[_account][_lp].balance) revert InsufficientFunds();
+        uint256 previousBalance = _accountToLPData[_account][_lp].balance;
+        if (_amount > previousBalance) revert InsufficientFunds();
         _syncDelegation(_account);
 
         /// @dev ensure proper accounting in case of admin error in `modifyScalingFactor` where not all holders are updated
         _syncVotingWeight(_account, _lp);
-        _accountToLPData[_account][_lp].balance -= _amount;
 
-        _burn(_account, _amount * scalingFactors[_lp] / FIXED_POINT_PERCENT);
+        /// @dev `_syncVotingWeight` only adjusts the stored scaling factor, so `previousBalance` is still current
+        uint256 newBalance = previousBalance - _amount;
+        _accountToLPData[_account][_lp].balance = newBalance;
+
+        _burnUpToBalance(_account, _scaledBalance(previousBalance, _lp) - _scaledBalance(newBalance, _lp));
+
         bool success = IERC20(_lp).transfer(_account, _amount);
         if (!success) revert TransferFailed();
 
@@ -201,18 +209,33 @@ contract ButteredBread is IButteredBread, ERC20VotesUpgradeable, Ownable2StepUpg
             _accountToLPData[_account][_lp].scalingFactor = currentScalingFactor;
 
             if (lpBalance > 0) {
-                if (currentScalingFactor > initialScalingFactor) {
-                    _mint(
-                        _account,
-                        (lpBalance * currentScalingFactor - lpBalance * initialScalingFactor) / FIXED_POINT_PERCENT
-                    );
+                /// @dev difference of scaled balances, matching how `_deposit` and `_withdraw` scale
+                uint256 currentWeight = lpBalance * currentScalingFactor / FIXED_POINT_PERCENT;
+                uint256 initialWeight = lpBalance * initialScalingFactor / FIXED_POINT_PERCENT;
+
+                if (currentWeight > initialWeight) {
+                    _mint(_account, currentWeight - initialWeight);
                 } else {
-                    _burn(
-                        _account,
-                        (lpBalance * initialScalingFactor - lpBalance * currentScalingFactor) / FIXED_POINT_PERCENT
-                    );
+                    _burnUpToBalance(_account, initialWeight - currentWeight);
                 }
             }
         }
+    }
+
+    /// @notice Apply the scaling factor of a liquidity pool to an LP token balance
+    function _scaledBalance(uint256 _balance, address _lp) internal view returns (uint256) {
+        return _balance * scalingFactors[_lp] / FIXED_POINT_PERCENT;
+    }
+
+    /**
+     * @notice Burn `ButteredBread`, capped at the balance of the account
+     * @dev Accounts that deposited before scaled balances were differenced hold slightly less `ButteredBread`
+     *  than their aggregate entitlement, because each deposit floored its own mint. Capping lets those accounts
+     *  withdraw their full LP balance instead of reverting in `_burn`. For every deposit made under the current
+     *  accounting the requested amount is already covered, so the cap does not apply.
+     */
+    function _burnUpToBalance(address _account, uint256 _amount) internal {
+        uint256 accountBalance = balanceOf(_account);
+        _burn(_account, _amount > accountBalance ? accountBalance : _amount);
     }
 }
--- a/test/ButteredBread.t.sol
+++ b/test/ButteredBread.t.sol
@@ -377,11 +377,12 @@ contract ButteredBreadTest_Fuzz is ButteredBreadTest {
         assertEq(bb.accountToLPBalance(ALICE, GNOSIS_CURVE_POOL_XDAI_BREAD), _s.deposit);
         assertEq(curvePoolXdai.balanceOf(ALICE), 0);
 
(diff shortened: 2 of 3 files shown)


BreadrichEngels (approved):
## Review — `fix: burn the difference of scaled balances so full LP withdrawals clear`

Verified the diff against the full `ButteredBread.sol` at `e82eea5`, not just the description. The root-cause analysis is correct and the fix is minimal, well-reasoned, and thoroughly tested. **Approving.** Two notes below are non-blocking, plus a process ask given this upgrades a live funds-holding proxy.

### The fix is correct

The bug is exactly as described: `_deposit` floored per-deposit (`Σ floor(aᵢ·f/P)`) while `_withdraw` floored the aggregate (`floor(Σaᵢ·f/P)`), and `Σ floor ≤ floor Σ`, so a full exit tried to burn up to a few wei more `BB` than was ever minted → `ERC20InsufficientBalance`.

Moving both sides to the **difference of scaled balances** is the right shape, and it telescopes:
- Deposits: `Σ [floor(bₙ·f/P) − floor(bₙ₋₁·f/P)]` collapses to `floor(b_total·f/P)`, so cumulative minted supply == aggregate entitlement (your `testMintedSupplyMatchesAggregateEntitlement` pins this).
- A full withdraw burns `floor(prev·f/P) − 0 = floor(prev·f/P)`, exactly the minted amount → clears to zero.

I also confirmed the `_syncVotingWeight` change is a genuine latent-bug fix, not just cosmetic: the old `(lp·current − lp·initial)/P` is `floor` of a product-difference, whereas the new `lp·current/P − lp·initial/P` is the difference-of-floors that keeps a synced holding at exactly `_scaledBalance(balance)`. The old form could drift ±1 wei from the invariant every re-scale. Good catch folding it in.

`_scaledBalance` and `_burnUpToBalance` are the only new code paths, and I checked there are no other `_mint`/`_burn` call sites left on the old `_amount * factor / P` arithmetic — all three are converted.

### Strengths

- **Layout-safe upgrade** confirmed — no new state; the two helpers are `view`/internal. `bread`, `allowlistedLPs`, `scalingFactors`, `_accountToLPData` keep their slots.
- **Test coverage is excellent** — the incident-replay at the live factor `32` with the exact on-chain amounts, the `132` case proving it's not sub-100-specific, the fork+`upgradeAndCall` test asserting all three affected accounts revert before / fully exit after, and the state-preservation test. This is the standard I'd want on a funds-touching accounting change.
- Changing `testWithdrawAndBurn` to assert `balanceOf == remaining·f/P` (the invariant) rather than the old delta is strictly stronger — agreed.

### Notes (non-blocking)

1. **Legacy partial-withdrawal dust corner.** `_burnUpToBalance` cleanly enables full exits for the pre-fix shortfall accounts. But for a *legacy* account doing a near-full **partial** withdrawal that leaves a remaining scaled balance smaller than its (few-wei) shortfall, the cap bites and zeroes their holding while a sub-wei-of-`BB` LP dust is still staked — i.e. voting weight drops to 0 slightly early on that dust. Economically negligible and they can still withdraw the remainder, but note the fuzz tests (`testFuzzPartialWithdrawalsKeepBalancesInSync`) only exercise *new-accounting* positions, so this exact legacy-partial path is uncovered. A one-off regression over a seeded-shortfall account would close the gap if you want it airtight.

2. **`_burnUpToBalance` silently absorbs any over-burn.** For positions minted under the new accounting the cap provably never applies (holding == entitlement), so it's a pure safety net for legacy state — the right call. The tradeoff is that a *future* accounting regression that over-burns a healthy position would be swallowed rather than reverting loudly. The doc-comment already scopes intent; just flagging that the invariant `_amount ≤ balanceOf` for new positions now lives only in tests, not an assert.

### Process

You correctly deferred the two policy calls — (a) whether `32` is the intended factor (0.32× contradicts the documented `100 = 1X` semantics; and `initialize` lacks the `≥ FIXED_POINT_PERCENT` guard that `modifyScalingFactor` enforces, so prod sits at a value it can't be reset to) and (b) `onlyAllowed` gating `withdraw`, which would freeze positions in a delisted LP. Both are real and worth their own issues; the allowlist-gates-withdraw one in particular is a latent lockup risk. Please file those so they don't get lost with this PR.

Since this is still a **draft** and it upgrades a live Gnosis proxy holding user funds, I'd want a second domain set of eyes (Ron) on the upgrade mechanics specifically — `ProxyAdmin` owner path in `upgradeAndCall`, and confirming the mainnet upgrade will be driven the same way the fork test drives it — before the on-chain upgrade is actually executed. The Solidity change itself I'm confident in.

Nice work — the writeup, the live-account reproduction, and the recovery-without-upgrade recipe are all model-grade.

